PR #1, opened 2020-02-26
Blocks can be smushed through other blocks by pressing 'down'

Poobslag:
Tapping 'down' on a block will try to smush it vertically through other
blocks. Holding 'down' will smush a block at the last moment. Tapping
'down' can also prevent a piece from locking, although this cannot be
used indefinitely.

Refactored block movement logic to use state pattern

There's now a brief delay after a piece locks before lines are cleared.
This delay has been subtracted from the 'appearance delay'. The primary
purpose of this delay was to facilitate 'piece smushing', although
there's also an arguable aesthetic benefit to seeing a piece dropped
before lines are cleared.


scenes/Piece.gd
```diff
@@ -236,49 +223,111 @@ func increment_input_frames(frames, action):
...
+			perform_lock_reset()
+
+func perform_lock_reset():
+	if piece.lock_resets >= MAX_LOCK_RESETS:
+		return false
+	piece.lock = 0
+	piece.lock_resets += 1
+	return true
+
+func calc_smush_target():
```
Poobslag: Add some method comments; these methods are sort of confusing by nature

PR #6, opened 2020-03-03
Code cleanup: private methods, static typing

Poobslag:
Prefixed private methods, fields with underscores. I didn't prefix
private constants: all constants are essentially private anyways because
it's such a pain in the ass to reference them from other classes. I also
didn't prefix fields like 'NextPieces' which act as outside references.

Signals now use past-tense names, 'game started', 'line cleared'.

Variables, functions include static type hints.


scenes/CornerMap.gd
```diff
@@ -4,41 +4,41 @@ Fine-grained tilemap which draws the inner corners of 90-degree angles.
...
 
 # constants used when drawing blocks which are connected to other blocks
-const CONNECTED_UP = 1
-const CONNECTED_DOWN = 2
-const CONNECTED_LEFT = 4
-const CONNECTED_RIGHT = 8
+const _CONNECTED_UP = 1
+const _CONNECTED_DOWN = 2
+const _CONNECTED_LEFT = 4
+const _CONNECTED_RIGHT = 8
```
Poobslag: Don't prefix these constants.

scenes/CornerMap.gd
```diff
@@ -4,41 +4,41 @@ Fine-grained tilemap which draws the inner corners of 90-degree angles.
...
 
-onready var ParentMap = get_node("..")
+onready var _ParentMap = get_node("..")
```
Poobslag: Don't prefix this outside reference.

scenes/CornerMap.gd
Poobslag: Fixed; constants are no longer prefixed.

scenes/CornerMap.gd
Poobslag: Fixed; references are no longer prefixed.

PR #8, opened 2020-03-06
Added 'practice menu' screen with different marathon difficulties.

Poobslag:


scenes/Piece.gd
```diff
@@ -241,6 +248,7 @@ func _apply_player_input() -> void:
 		return
 	
 	var did_hard_drop := false
+	var did_smush := false
```
Poobslag: Unused local variable

Poobslag: Removed

PR #11, opened 2020-03-07
Polished up sprint, ultra modes. 'Finish' instead of 'win'.

Poobslag:
Finishing sprint/ultra now plays a general 'finished' noise instead of a
'you win' sound. If you get an A or better grade, there is also
applause. This applause also applies to marathon.

Holding the rotate key now spawns a piece consistently at the top of the
well. Some pieces were spawning one space lower because of how they
rotate.

Tests now more broadly assert the grade performance (S, A+) instead of
an exact rank. The 'exact rank' tests were too fiddly.


scenes/RankCalculator.gd
```diff
@@ -87,6 +89,10 @@ const RDF_LINES := 0.96
 const RDF_BOX_SCORE := 0.96
 const RDF_COMBO_SCORE := 0.98
 
+const MASTER_BOX_SCORE := 13.6
```
Poobslag: Document these new constants

scenes/PieceTypes.gd
```diff
@@ -22,10 +22,17 @@ const KICKS_P = [
 	]
 
 const KICKS_V = [
-		[Vector2(-1,  0), Vector2(-1, -1), Vector2( 0, -1), Vector2( 0,  2), Vector2( 0, -2)],
```
Poobslag: Why did these kicks change? Document it in the commit message

PR #56, opened 2020-03-27
Replaced assets with large 1024x1024 versions.

Poobslag:
These 1024x1024 images use a limited color palette so we can recolor
customer sprites.


art/eat-anim/head.png.import
```diff
@@ -21,7 +21,7 @@ compress/bptc_ldr=0
 compress/normal_map=0
 flags/repeat=0
 flags/filter=true
-flags/mipmaps=false
+flags/mipmaps=true
```
Poobslag: Were mipmaps enabled deliberately?

Poobslag: I've removed this file; it was redundant with /art/customer/head.png

PR #57, opened 2020-03-27
Harder difficulties start out locked.

Poobslag:
This isn't for gatekeeping, but just because I don't want players
jumping into the harder difficulties and having an unfun time or missing
the point of the game. An A- grade is pretty low; you can get this on
Normal Marathon by only clearing 20-30 lines, for example.

### DIFF
--- a/scenes/Global.gd
+++ b/scenes/Global.gd
@@ -96,10 +96,13 @@ func grade(rank: float) -> String:
 	elif rank <= 48: return "D+"
 	elif rank <= 58: return "D"
 	elif rank <= 64: return "D-"
-	else: return "-"
+	else: return NO_GRADE
 
 # How long it takes a customer to grow to a new size
 const CUSTOMER_GROWTH_SECONDS := 0.12
 
 # How large customers can grow; 5.0 = 5x normal size
 const MAX_FATNESS := 10.0
+
+# String to display if the player scored worse than the lowest grade
+const NO_GRADE := "-"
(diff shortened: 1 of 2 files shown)


scenes/Global.gd
```diff
@@ -96,10 +96,13 @@ func grade(rank: float) -> String:
...
-	else: return "-"
+	else: return NO_GRADE
 
 # How long it takes a customer to grow to a new size
 const CUSTOMER_GROWTH_SECONDS := 0.12
 
 # How large customers can grow; 5.0 = 5x normal size
 const MAX_FATNESS := 10.0
+
+# String to display if the player got below the minimum grade
```
Poobslag: "Minimum grade" for what?

scenes/PracticeMenu.gd
```diff
@@ -18,60 +21,110 @@ func _ready() -> void:
...
+				best_grade = Global.grade(rank_result.score_rank)
 	if best_score == 0:
 		return ""
 	if has_lived:
 		return "Top: %s (%s)*" % [best_score, best_grade]
 	else:
 		return "Top: %s (%s)" % [best_score, best_grade]
 
+"""
+Calculates the best rank a player's gotten for a specific sprint scenario.
```
Poobslag: Grammar; maybe something more formal like "received" instead of "gotten"

scenes/Global.gd
Poobslag: Rephrased; "scored worse than the lowest grade"

scenes/PracticeMenu.gd
Poobslag: Rephrased; "best rank a player's received"

PR #58, opened 2020-03-28
Customers now appear in 16 different colors.

Poobslag:
The current approach uses a complex 'RgbPalette.shader' shader, which
relies on an unusual style of sprite. The unusual sprite involves three
red, green and blue channels which each include data for custom colors,
and the remaining 'black' channel is used for lines. The shader converts
these red/green/blue colors to arbitrary colors, e.g aquamarine, puce
and taupe.

If, in the future, more than 3 colors are needed, this shader can be
applied to multiple sprites layered on top of one another. Hopefully
that's sufficient, because even as kludgy as this current approach is,
it involved a lot of trial and error, and about 2-3 failed technical
approaches. The benefits of the current approach are that the heavy
lifting can be offloaded to graphics cards. The drawbacks are that the
sprite files need to be ugly images with bright colors.

One failed approach was a naive shader-based approach which recolored the existing sprites based on whether they matched certain arbitrary colors (e.g change forest green to orange.) This failed because the shader was applied after sprites were scaled, so the user could clearly see the aliased pixels in the resulting algorithm.

Another failed approach involved loading the texture images and
replacing the pixels one-by-one. This looked good but was exceptionally
slow, taking about 15 seconds to recolor a customer even with the
current limited number of animations. The mouth-sheet.png files is
8215x1026, for example, and computers are not vet fpushry good at doing
something 8,000,000 times.


scenes/Restaurant/Customer.gd
```diff
@@ -57,6 +57,63 @@ func feed() -> void:
...
 		show_food_effects(0.066)
 
+"""
+Recolors the customer according to the specified color definition. This involves updating shaders and sprite
+properties.
+
+Parameter: 'color_def' describes the sprites to recolor and how to recolor them.
+"""
+func recolor(color_def: Dictionary) -> void:
+	var line_color := Color("6c4331")
```
Poobslag: if color_def lacks a "line" property, these various 'black' shader properties will retain their previous values. Nothing to fix necessarily, but could pose strange problems in the future.

scenes/RgbPalette.shader
```diff
@@ -0,0 +1,22 @@
+shader_type canvas_item;
+
+// the replacement colors
+uniform vec4 red : hint_color;
+uniform vec4 green : hint_color;
+uniform vec4 blue : hint_color;
+uniform vec4 black : hint_color;
+
+void fragment() {
```
Poobslag: Add a class-level description of what this shader is all about. It's crazy.

Poobslag: Added verbose description. I agree, it is quite crazy. Maybe we should publish this on the godot forums.

PR #65, opened 2020-03-29
Reduced load times. Customer resources are now loaded in a thread.

Poobslag:
Customer resources are handled by the new CustomerLoader class. This
class loads resources in a background thread so the game launches
in less than a second. However, customers don't show up until the
textures are loaded. This is acceptable as thematically, they're living
beings who need to sit down. It's not like having a lamp randomly appear.

Seat.tscn no longer has a customer visible by default. This is because
having a customer visible slowed down the game was loading, since that
customer's resources would need to be loaded just to be thrown away.
Customer.gd includes a 'customer_preset' property which can be toggled
to load or unload a default customer. This requires the 'tool' keyword
to enabled for the customer class, which can cause the editor to go
crazy or crash. So, this functionality is disabled by default.

Added a benchmark tool.

Removed some unnecessary print statements when creating boxes, kicking
pieces. The 'piece kick' messages seemed helpful so they can be enabled
with a 'debug' field.

Moved some customer-related constants from Global into the new CustomerLoader
class.


scenes/Global.gd
```diff
@@ -98,30 +107,19 @@ func grade(rank: float) -> String:
...
-# How large customers can grow; 5.0 = 5x normal size
-const MAX_FATNESS := 10.0
-
-# String to display if the player scored worse than the lowest grade
-const NO_GRADE := "-"
+"""
+Sets the start time for a benchmark. Calling 'benchmark_start(foo)' and 'benchmark_finish(foo)' will display a message
+like 'foo took 123 milliseconds'.
+"""
+func benchmark_start(var key := ""):
```
Poobslag: ` -> void`

scenes/Global.gd
```diff
@@ -98,30 +107,19 @@ func grade(rank: float) -> String:
...
-	{"line_rgb": "6c4331", "body_rgb": "f9f7d9", "eye_rgb": "91e6ff ffffff", "horn_rgb": "ffffff"}, # white
-	{"line_rgb": "3c3c3d", "body_rgb": "1a1a1e", "eye_rgb": "b8260b f45e40", "horn_rgb": "282828"}, # black
-	{"line_rgb": "6c4331", "body_rgb": "7a8289", "eye_rgb": "f5f0d1 ffffff", "horn_rgb": "282828"}, # grey
-	{"line_rgb": "41281e", "body_rgb": "0b45a6", "eye_rgb": "fad541 ffffff", "horn_rgb": "282828"}  # dark blue
-]
+"""
+Prints the amount of time which has passed since a benchmark was started. Calling 'benchmark_start(foo)' and
+'benchmark_finish(foo)' will display a message like 'foo took 123 milliseconds'.
+"""
+func benchmark_end(var key := ""):
```
Poobslag: ` -> void`

scenes/Global.gd
Poobslag: Additionally, `var` is redundant here

scenes/Global.gd
Poobslag: Additionally, `var` is redundant here

scenes/Restaurant/CustomerLoader.gd
```diff
@@ -0,0 +1,323 @@
...
+
+Parameter: 'customer_def' is a customer definition which has been previously passed to the 'summon_customer' method.
+"""
+func is_customer_ready(customer_def: Dictionary) -> bool:
+	return _ready_customer_definitions.erase(customer_def)
+
+"""
+Waits for all threads to finish. Threads must be disposed (or 'joined') for portability.
+"""
+func wait_to_finish():
```
Poobslag: Additionally, `var` is redundant here

scenes/Restaurant/CustomerLoader.gd
```diff
@@ -0,0 +1,323 @@
...
+	for thread in _threads:
+		thread.wait_to_finish()
+
+func _exit_tree():
```
Poobslag: Additionally, `var` is redundant here

scenes/Restaurant/CustomerLoader.gd
```diff
@@ -0,0 +1,323 @@
...
+]
+
+# pool of threads which are currently loading resources
+var _threads := [];
+
+# list of customer definitions which are loaded, but haven't been asked about
+var _ready_customer_definitions := {}
+
+# mutex which ensures we don't load multiple resources concurrently
+var _load_mutex = Mutex.new()
```
Poobslag: strong typing; `:=` instead of `=`

scenes/Restaurant/CustomerLoader.gd
```diff
@@ -0,0 +1,323 @@
...
+		thread.wait_to_finish()
+
+func _exit_tree():
+	# wait until all threads are done to dispose of them
+	wait_to_finish()
+
+"""
+Prunes terminated threads from our thread pool.
+"""
+func _prune_unused_threads():
```
Poobslag: `unused` and `terminated` are both slightly inaccurate, and it would be better just to say `inactive` since that's what the thread method is called

scenes/Restaurant/CustomerLoader.gd
```diff
@@ -0,0 +1,323 @@
...
+	var far_leg_outline: Texture
+	var near_leg_outline: Texture
+	var near_arm_outline: Texture
+	var head_outline: Texture
+	var far_arm: Texture
+	var near_arm: Texture
+	var near_leg: Texture
+	var far_leg: Texture
+	var head: Texture
+	if customer_def.has("body"):
```
Poobslag: It's a little strange to check for a property which is never used. This warrants a code comment.

Poobslag:
Your commit comment should mention that you moved all the `eat-anim` resources to a new folder.

scenes/Restaurant/CustomerLoader.gd
Poobslag: While there was a typo in your comment, I've added the missing `-> void`;

scenes/Restaurant/CustomerLoader.gd
Poobslag: While there was a typo in your comment, I've added the missing `-> void`;

scenes/Restaurant/CustomerLoader.gd
Poobslag: Changed to `:=`

scenes/Restaurant/CustomerLoader.gd
Poobslag: Changed to `prune_inactive_threads`. Also changed comment.

scenes/Restaurant/CustomerLoader.gd
Poobslag: Added a comment here as well as the `summon_customer` method describing this functionality

PR #66, opened 2020-03-30
Customers make sounds when they enter, eat, and leave.

Poobslag:
When they enter, a door chimes and sometimes they say 'hello'. If the
player gets a big combo, the customer will says 'mmm!' or 'oooh!' in between
bites. When the player breaks their combo, the customer says 'check please!"

The customer's sounds are AudioStreamPlayer2Ds with a spatial element.
The door is on the left side of the store, so when the door chimes or
customers say 'hello', the sounds come from the left. When customers say
'mmm!' or 'check please!' then the sound comes from close to the center.

I also tried to make the other customer sounds lean a little to the left, since
the customers are on the left side of the screen. I figure kitchen sounds will
be panned right and eating sounds will be panned left.

There's some logic to make sure customers don't say 'hello' 'goodbye'
too frequently.


Poobslag:
Additionally, these sounds replace most of the chef's sounds. The chef still speaks at the beginning of a round, and when she loses. I didn't think it made thematic sense for the chef to be chattering about how great she is while she cooks. Let the customers say that stuff.

scenes/Game.gd
```diff
@@ -102,12 +102,16 @@ func end_game(delay: float, message: String) -> void:
 Relays the 'line_cleared' signal to any listeners, and triggers the 'customer feeding' animation
 """
 func _on_line_cleared(lines_cleared: int) -> void:
+	var customer_talks: bool = $Playfield.combo >= 5 && lines_cleared > ($Playfield.combo + 1) % 3
```
Poobslag: This line of code is pretty obscure, and warrants a code comment

Poobslag: Added a comment.

PR #98, opened 2020-04-02
Fixed stool shadows. Added procedural outlines.

Poobslag:
Stools now darken when a customer sits on them, and lighten when
customers stand up.

Outlines are generated procedurally instead of being hand-drawn.

Some objects can use a simple Outline.shader but the customer object is
more complex, and needs a MultiOutline.shader and MultiOutlineSprite for
now. I'll create a technical debt defect to track this future work.


scenes/MultiOutline.shader
```diff
@@ -0,0 +1,31 @@
+/*
+ * Adapted from outline.shader obtained from GDQuest's godot-demos, 2020-04-01
```
Poobslag, line 2: Add a better description to this shader.

scenes/Outline.shader
```diff
@@ -0,0 +1,32 @@
+/*
+ * Adapted from outline.shader obtained from GDQuest's godot-demos, 2020-04-01
```
Poobslag, line 2: Add a better description for this shader.

scenes/MultiOutline.shader
Poobslag: Improved description, describing in detail why there's a "multi outline" shader and how it differs from the regular outline shader.

scenes/Outline.shader
Poobslag: Improved description, describing in detail why there's a "multi outline" shader and how it differs from the regular outline shader.

PR #106, opened 2020-04-05
Node-based state pattern for pieces. Minor refactoring.

Poobslag:
PieceManager already used a function-based state pattern where it stored
the name of a function to call. It now uses a node-based state pattern.
The state machine code is generic, and has been extracted to
/src/main/state-machine.gd to encourage reuse.

Merged PieceManager.tscn into Puzzle.tscn. PieceManager could not
function without a source of random pieces and playfield constraints. It
had logic for providing its own if these dependencies were null, but
that kind of code seemed out-of-place in src/main.

ShadowMap no longer requires a reference to '../TileMap', it now provides
shadows for its parent.

Fixed infinite loop when dropping an empty piece.


src/main/state.gd
```diff
@@ -0,0 +1,11 @@
+class_name State
+extends Node
+
+var frames := 0
+
+func enter(host) -> void:
```
Poobslag, line 6: Add documentation for these interface methods.

src/main/state.gd
```diff
@@ -0,0 +1,11 @@
+class_name State
+extends Node
```
Poobslag, line 2: Add class-level documentation

src/main/state.gd
```diff
@@ -0,0 +1,11 @@
...
+
+var frames := 0
```
Poobslag, line 4: Document this field.

src/main/state.gd
```diff
@@ -0,0 +1,11 @@
...
+
+func enter(host) -> void:
+	pass
+
+
+func update(host) -> String:
```
Poobslag, line 10: Static type these host fields -- they have to be "Nodes" if nothing else, because of this class's design

src/main/state.gd
Poobslag: Added documentation.

src/main/state.gd
Poobslag: Documented the field.

src/main/state.gd
Poobslag: Added static typing for Node.

src/main/state.gd
Poobslag: Added class-level documentation.

src/main/state.gd
Poobslag: This is not a good idea, because static typing the host fields prevents implementors from providing their own types. I'll leave them untyped.

PR #111, opened 2020-04-08
Fixed puzzle freezing when prelock was interrupted.

Poobslag:
The prelock state can result in two states; writing the piece to the
playfield or going back to the move state if the player smushes or
performs a lock reset. If these two state changes overlapped, bugs
resulted such as infinite loops or freezes.

I've refactored the surrounding code to minimize the number of places
where we manage the piece state using 'set_state'. The states themselves
now monitor 'apply_player_input' and change the state themselves.

### DIFF
--- a/src/main/puzzle/piece/piece-manager.gd
+++ b/src/main/puzzle/piece/piece-manager.gd
@@ -20,8 +20,8 @@ var _input_ccw_frames := 0
 
 var _gravity_delay_frames := 0
 
-onready var playfield = $"../Playfield"
-onready var _next_pieces = $"../NextPieces"
+onready var playfield: Playfield = $"../Playfield"
+onready var _next_pieces: NextPieces = $"../NextPieces"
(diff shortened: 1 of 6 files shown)


src/main/puzzle/piece/piece-manager.gd
```diff
@@ -80,15 +78,26 @@ func _physics_process(_delta: float) -> void:
...
+
+Returns true if the newly written piece results in a pause of some sort.
+"""
+func write_piece_to_playfield() -> bool:
+	var result := playfield.write_piece(active_piece.pos, active_piece.orientation, active_piece.type,
+			piece_speed.line_clear_delay)
+	clear_piece()
+	return result
+
+
```
Poobslag, line 96: This comment is inaccurate.

Poobslag: The comment is accurate, Github was just displaying the surrounding code in a misleading way.

PR #114, opened 2020-04-10
Added FPS tracker; tweaked gravity physics

Poobslag:
The FPS tracker can be enabled by typing 'BIGFPS' in the overworld. It
will most likely be capped at 60, although you can temporarily uncap the FPS by
disabling VSYNC in the project settings.

I encountered a transient error where landing on a platform caused the
player to constantly bounce up and down, with their Y value oscillating between
-4 and 0. I'm changing the gravity code to reset their Y velocity as a
countermeasure, although this bug is difficult to reproduce.

### DIFF
--- a/fps-label.gd
+++ b/fps-label.gd
@@ -1,0 +1,7 @@
+extends Label
+"""
+Renders the current frames-per-second to the screen, '60.00 fps'
+"""
+
+func _process(delta: float):
+	text = "%.2f fps" % Engine.get_frames_per_second()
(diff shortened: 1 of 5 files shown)


fps-label.gd
```diff
@@ -0,0 +1,4 @@
+extends Label
```
Poobslag, line 1: Add documentation

Poobslag: Added documentation

PR #120, opened 2020-04-15
Animated backwards-facing customers.

Poobslag:
Customer spritesheets now follow a convention where frame 0 is empty, frame 1
faces southeast, and frame 2 faces northwest.

Increased size of customer's 3D hitbox, to avoid clipping walls and terrain.
Some clips still occur now and then, but that's OK. As long as they're not all
the time.

Added protection around set_frame call in MultiOutlineSprite to avoid logging errors

Removed unused 'NearMovement' sprite.

Fixed some minor issues with jump animations. Jumping from a standing
start did not fire an animation, now it does. Holding the jump button to
bounce did not trigger animations, now it does.


src/main/world/restaurant/customer.gd
```diff
@@ -374,9 +532,9 @@ func _update_customer_properties() -> void:
 	
 	if Engine.is_editor_hint():
 		# Skip the sound effects if we're using this as an editor tool
-		return
```
Poobslag, line 377: In compliance with GDQuest's code style guidelines, I've removed the 'return' from the middle of this statement

src/main/world/shadow-mesh.gd
```diff
@@ -0,0 +1,16 @@
+extends MeshInstance
+"""
+A mesh which casts an appropriate shadow based on the customer's position and status.
+"""
+
+const DEFAULT_SHADOW_SCALE = Vector3(0.65, 0.85, 0.5)
```
Poobslag, line 6: `:=` type hint? Or can you use `: Vector3`

src/main/world/shadow-mesh.gd
```diff
@@ -0,0 +1,16 @@
...
+
+# a scalar which makes the customer's shadow bigger/smaller
+export (float) var _shadow_scale := 1.0 setget set_shadow_scale
+
+"""
+Sets a scalar which makes the customer's shadow bigger/smaller.
+"""
+func set_shadow_scale(var shadow_scale) -> void:
```
Poobslag, line 14: Type hint for `shadow_scale`

src/main/world/shadow-mesh.gd
Poobslag: Added missing type hint

src/main/world/shadow-mesh.gd
Poobslag: Added missing type hint

PR #124, opened 2020-04-15
Changed window scaling. Resolved console errors. Hid FPS.

Poobslag:
Window scaling now preserves the aspect ratio. This aspect ratio skewing wasn't
a big deal for 2D but looks really bad in 3D.

Disabled font oversampling. This prevents some warnings, 'Font
oversampling does not work in 'viewport' stretch mode, only 2D.

Fixed json parse error which was caused by attempting to parse a blank line.

Fixed 'ViewportTexture: Path to node is invalid' which seems to be
caused by having the Viewport lower in the scene tree than the
ViewportTexture. (Godot issue #27790)


src/main/puzzle/scenario-history.gd
```diff
@@ -114,11 +114,12 @@ func load_scenario_history() -> void:
 	if save_game.file_exists(scenario_history_filename):
 		save_game.open(scenario_history_filename, File.READ)
 		while not save_game.eof_reached():
-			var current_line = parse_json(save_game.get_line())
-			if current_line == null:
+			var line_string = save_game.get_line()
```
Poobslag, line 117: Type hints

Poobslag: Added missing type hints

PR #127, opened 2020-04-15
Turbo slips from ledges in 3D mode.

Poobslag:
This prevents her from hanging off a ledge in mid-air, which was jarring
for two reasons: she was sort of suspended in mid-air megaman style with
maybe a toe touching the ledge or maybe not, which looked strange. But also, there was a visible
gap between her shadow and the ledge's shadow.

I also refactored a few references to use Vector2.ZERO and Vector3.ZERO
for readability.


src/main/world/turbo.gd
```diff
@@ -36,8 +39,14 @@ const CAMERA_LEAD_DISTANCE := 60
 # Number from [0.0, 1.0] for how tight the camera should snap to Turbo's movement
 const CAMERA_JERKINESS := 0.04
 
+# How many rays we should cast when determining how Turbo falls from a cliff
```
Poobslag, line 42: Use "ledge" instead of "cliff" for consistency

src/main/world/turbo.gd
```diff
@@ -36,8 +39,14 @@ const CAMERA_LEAD_DISTANCE := 60
...
+const LEDGE_RAY_COUNT = 16
+
+# How far apart the rays should be when determining how Turbo falls from a cliff
+const LEDGE_RAY_RADIUS = 12.0
```
Poobslag, line 46: `:=`

src/main/world/turbo.gd
```diff
@@ -36,8 +39,14 @@ const CAMERA_LEAD_DISTANCE := 60
 # Number from [0.0, 1.0] for how tight the camera should snap to Turbo's movement
 const CAMERA_JERKINESS := 0.04
 
+# How many rays we should cast when determining how Turbo falls from a cliff
+const LEDGE_RAY_COUNT = 16
```
Poobslag, line 43: `:=`

src/main/world/turbo.gd
```diff
@@ -216,26 +229,50 @@ func jump() -> void:
...
+			if collision.collider.get("foothold_radius") and translation.y > collision.collider.translation.y + 0.2:
+				# Turbo is standing on something she might slip from
+				var slip_velocity: Vector3 = translation - collision.collider.translation
+				slip_velocity.y = 0
+				if slip_velocity.length() > collision.collider.foothold_radius:
+					# Turbo is too far from the center, and should slip
+					just_slipped = true
+					accelerate_player_xy(delta, Vector2(slip_velocity.x, slip_velocity.z), MAX_SLIP_ACCELERATION, MAX_NARROW_SLIP_SPEED)
+					$CoyoteTimer.stop()
+					$JumpBuffer.stop()
```
Poobslag, line 251: This `CoyoteTimer.stop JumpBuffer.stop` code should probably be pulled out, as it likely applies to slipping from ledges as well. Both or neither.

src/main/world/turbo.gd
Poobslag: Changed to 'ledge'

src/main/world/turbo.gd
Poobslag: Changed to `:=`

src/main/world/turbo.gd
Poobslag: Removed `CoyoteTimer.stop` `JumpBuffer.stop` code -- i've added an item to the backlog to consider what the CoyoteTimer should do now. Maybe the player needs to be able to jump while slipping in some cases.

PR #131, opened 2020-04-17
Added conversation font, animation. Added enum type hints.

Poobslag:
The animation supports pauses. The text will pause at newlines, and will
pause for slash characters which are hidden from the user.

I considered something more automated like pausing at periods, and
experimented with pausing at words too. It didn't feel very smooth, and
pausing at periods runs into edge cases in sentences like "H. G. Wells
paid $0.30 for a No. 2 pencil."

The text speed should eventually be customizable via a settings menu. 2.0
is a good default speed. 1.0 matches spoken speech. There is no upper
limit.

Enabled font oversampling. I'm aware this causes warnings in some scenes,
but it makes the dialog text much smoother when increasing window size.


src/main/ui/conversation-control.gd
```diff
@@ -0,0 +1,10 @@
+extends Control
+"""
+Displays UI elements for having a conversation.
+"""
+
+"""
+Starts a new conversation.
```
Poobslag, line 7: Use the other comment, as it's more descriptive: `Animates the conversation label to gradually reveal the specified text.`

src/main/ui/conversation-label.gd
```diff
@@ -0,0 +1,60 @@
...
+var DEFAULT_GLYPH_DELAY := 0.05
+
+# How many seconds to delay when displaying whitespace or the special pause character, '/'.
+var SPECIAL_GLYPH_DELAYS := {
+	" ": 0.00,
+	"/": 0.40,
+	"\n": 0.80,
+}
+
+# Stores the amount of seconds to delay for each visible_characters index.
```
Poobslag, line 21: `Stores the delay in seconds for each...` is a little more succinct.

src/main/ui/conversation-control.gd
Poobslag: Improved comment.

src/main/ui/conversation-label.gd
Poobslag: Shortened comment.

PR #137, opened 2020-04-19
Added chat window sfx. Simplified ChatUi interface.

Poobslag:
Added sfx for when conversation windows appear, disappear, or when text
appears.

ChatUi can now be used with just "play_text" and "pop_out", you don't have
to explicitly call "pop_in".


src/main/ui/chat/chat-ui.gd
```diff
@@ -36,6 +46,9 @@ Also updates the chat UI's appearance based on the amount of text being displaye
 background texture.
 """
 func play_text(name: String, text: String, accent_def: Dictionary) -> void:
+	if not $ChatFrame/SentenceManager.label_showing():
```
Poobslag, line 49: For symmetry, consider adding a comment here. (Or delete the other one.)

Poobslag: Added a comment, `Ensure the chat window is showing...`

PR #140, opened 2020-04-21
Player can rewind/fast forward dialog

Poobslag:
Player can rewind dialog with 'shift'. They can view the previous 100 chat
messages, even ones from previous conversations.

Player can skip through dialog with 'Z', and can advance messages that are
appearing too slowly.

### DIFF
--- a/src/main/ui/chat/chat-ui.gd
+++ b/src/main/ui/chat/chat-ui.gd
@@ -7,20 +7,161 @@ signal chat_event_played
 
 signal pop_out_completed
 
-# array of ChatEvents the player is currently paging through
-var _chat_events: Array
+# how long the player needs to hold the button to skip all dialog
+const HOLD_TO_SKIP_DURATION := 0.6
 
-# index of the current ChatEvent being shown
-var _current_line := 0
+# array of ChatEvents the player can page through
+var _chat_events := []
+var _chat_event_index := 0
+
+# historical array of ChatEvents the player can rewind through
+var _prev_chat_events := []
+var _prev_chat_event_index := 0
+
+# how long the player has been holding the 'interact' button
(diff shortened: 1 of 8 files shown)


src/main/ui/chat/chat-ui.gd
```diff
@@ -7,20 +7,161 @@ signal chat_event_played
...
+			$ChatFrame.pop_out()
+
+
+"""
+Rewind to a previously shown dialog line.
+
+Returns 'true' if the rewind was successful, or 'false' if there were no lines to rewind to.
+"""
+func _decrement_line() -> bool:
+	var result := false
```
Poobslag, line 110: Maybe change `result` to something more self-documenting like `did_decrement`

src/main/ui/chat/chat-ui.gd
```diff
@@ -7,20 +7,161 @@ signal chat_event_played
...
+"""
+Advance to the next dialog line.
+
+This is most likely the next spoken dialog line if the player is advancing dialog normally. However, it can also be a
+historical dialog line if the player was rewinding dialog.
+
+Returns 'true' if the increment was successful, or 'false' if there were no more lines to show.
+"""
+func _increment_line() -> bool:
+	var result := false
```
Poobslag, line 135: Maybe change `result` to something more self-documenting like `did_increment`

src/main/ui/chat/chat-ui.gd
Poobslag: Changed to `did_decrement`

src/main/ui/chat/chat-ui.gd
Poobslag: Changed to `did_increment`

PR #146, opened 2020-04-25
Added backend dialog tree logic. Fixed console errors.

Poobslag:
JSON dialog structure now allows links. Branching dialog trees can be
parsed by ChatLibrary, navigated by ChatUi. ChatUi always just picks the 2nd
fork for now.

Resolved some console error messages. Customer3D's ViewportTexture
was reporting errors because the viewport was too low in the scene tree.
get_meta() calls were throwing errors for objects lacking metadata.

Pushed json chat event parsing logic into ChatEvent

More robust accent_defs implementation. Each customer defines their own accent_def which is pulled in by InteractableManager.

### DIFF
--- a/src/delint.sh
+++ b/src/delint.sh
@@ -12,5 +12,6 @@ grep -R -n "^			.\{108\}$" --include="*.gd" .
 grep -R -n "^				.\{104\}$" --include="*.gd" .
 grep -R -n "^					.\{100\}$" --include="*.gd" .
 
-# fields/variables missing type hint
-grep -R -n "var [^:]* = " --include="*.gd" .
+# fields/variables missing type hint. includes a list of whitelisted type hint omissions
+grep -R -n "var [^:]* = " --include="*.gd" . \
+  | grep -v "chat-library.gd:38"
--- a/src/main/ui/chat/chat-event.gd
+++ b/src/main/ui/chat/chat-event.gd
@@ -21,13 +21,19 @@ enum Mood {
 }
 
 # The name of the person speaking, or blank if nobody is speaking
-var name: String
+var who := ""
 
-# The text being spoken
+# The line of dialog
 var text: String
 
 # The behavior the chatter should perform while saying the text (laughing, sweating, etc)
 var mood: int = Mood.NONE
+
+# List of string keys corresponding to branches off of this chat event.
+var links: Array
+
+# List of dialog strings corresponding to branches off of this chat event.
+var link_texts: Array
 
 """
 The chat window changes its appearance based on who's talking. For example, one character's speech might be blue with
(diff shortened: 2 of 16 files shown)


src/main/ui/chat/chat-event.gd
```diff
@@ -41,3 +47,58 @@ appearance, such as 'blue', 'soccer balls' and 'giant'.
...
+		"cry1": m = Mood.CRY1
+		"sweat0": m = Mood.SWEAT0
+		"sweat1": m = Mood.SWEAT1
+		"rage0": m = Mood.RAGE0
+		"rage1": m = Mood.RAGE1
+		_: m = Mood.NONE
+	return m
+
+
+func _to_string():
```
Poobslag, line 103: `-> void`

src/delint.sh
```diff
@@ -12,5 +12,6 @@ grep -R -n "^			.\{108\}$" --include="*.gd" .
 grep -R -n "^				.\{104\}$" --include="*.gd" .
 grep -R -n "^					.\{100\}$" --include="*.gd" .
 
-# fields/variables missing type hint
-grep -R -n "var [^:]* = " --include="*.gd" .
+# fields/variables missing type hint. includes a list of whitelisted type hint omissions
+grep -R -n "var [^:]* = " --include="*.gd" . \
+  | grep -v "chat-library.gd:32"
```
Poobslag, line 17: this delinter reference is out of date, and results in a message

src/main/ui/chat/chat-library.gd
```diff
@@ -6,65 +6,42 @@ Dialog is stored as a set of json resources. This class parses those json resour
...
+	var json_tree: Dictionary
+	var parsed = parse_json(json)
+	if typeof(parsed) == TYPE_DICTIONARY and parsed.has(""):
+		json_tree = parsed
+	elif typeof(parsed) == TYPE_DICTIONARY:
+		# We were given a dictionary corresponding to a dialog line. Wrap it into a tree.
+		json_tree = {"" : [parsed]}
+	elif typeof(parsed) == TYPE_ARRAY:
+		# We were given an array corresponding to a dialog branch. Wrap it into a tree.
+		json_tree = {"" : parsed}
```
Poobslag, line 46: This could return something crazy like a float and a string; maybe expand this tree to report an error in those cases for symmetry with `ChatEvent.from_dict()`:

`push_error("Invalid json tree: " % json_link)`

src/main/ui/chat/chat-ui.gd
```diff
@@ -143,10 +135,13 @@ func _increment_line() -> bool:
...
-			# continue forward through chat events
-			_chat_event_index += 1
-			did_increment = true
+		var link_index := 0
+		var links: Array = _chat_tree.get_event().links
+		if links and links.size() >= 2:
+			print("Choose one: %s" % [_chat_tree.get_event().link_texts])
+			link_index = clamp(1, 0, links.size() - 1)
+			print("  (%s)" % _chat_tree.get_event().link_texts[link_index])
+		did_increment = _chat_tree.advance(link_index)
```
Poobslag, line 144: Split this into two branches for clarity: the 'navigating the default branch' branch, and the 'continuing' branch

src/main/ui/chat/chat-event.gd
Poobslag: `-> string` actually, but fixed

src/main/ui/chat/chat-library.gd
Poobslag: Now pushes an error.

src/main/ui/chat/chat-ui.gd
Poobslag: Split into three explicit branches instead of the two which were there before.

src/delint.sh
Poobslag: Resolved delinter warning.

src/main/ui/chat/chat-ui.gd
```diff
@@ -143,10 +135,20 @@ func _increment_line() -> bool:
...
-		if _chat_event_index + 1 < _chat_events.size():
-			# continue forward through chat events
-			_chat_event_index += 1
-			did_increment = true
+		var link_index: int
+		var links: Array = _chat_tree.get_event().links
+		if not links:
+			# No links are available, advance the dialog
+			pass
+		elif links and links.size() == 1:
```
Poobslag, line 143: `elif links` is redundant, you already checked the `not links` case

Poobslag: Removed `elif links` to only check size

PR #148, opened 2020-04-26
Puzzle timing changes; shorter pause when making boxes.

Poobslag:
Instead of pausing for 30 frames whether making a box or clearing a line, the
pauses are now vastly different. boxes only result in 50% of the delay of
a line clear. Overall this results in a much more frenetic feel to the
game. I increased the line clear delay to compensate but it might still be too
fast.

Added new 'DELAYS' cheat to see puzzle mode timing data.

Fixed bug where rank calculator could tell the player they scored an
impossible 22.5 points per line, if they finished a level with fewer than
8 line clears.


src/main/puzzle/playfield.gd
```diff
@@ -298,11 +291,22 @@ func _check_for_box(x: int, y: int, width: int, height: int, cake = false) -> bo
...
+			result = true
+	return result
+
+
 """
 Checks whether any lines in the playfield are full and should be cleared. Updates the combo, awards points, and plays
 sounds appropriately. Returns 'true' if any lines were cleared.
 """
-func _check_for_line_clear() -> bool:
+func _process_line_clear() -> void:
```
Poobslag, line 309: This comment is no longer accurate.

Poobslag: Fixed comment, and removed conditional code.

PR #150, opened 2020-04-29
Added rudimentary dialog choice UI. More work remains.

Poobslag:
Chat UI can display between 2-6 choices. Choices have little icons
indicating the mood of each choice. Eventually these choies may be
abbreviated; the player might choose 'I'm not sure' which leads to them
saying, 'I'm not sure, something might go wrong. I think we should try
something else' in a melancholy tone. These icons are meant to avoid the
case where the player picks a choice with an unintended tone.

Turbo is able to emote in response to the dialog as well. If Turbo is
sitting still she'll emote appropriately; if moving (or if she starts to
move) she'll stop emoting.

Moved chat demos into their own folder. Enhanced ChatUiDemo.

Removed filler dialog from ChatFrame; this was triggering events.

Added null check to ChatFrame. It's possible it'll receive an event before
the dialog tree begins, as was happening with the filler dialog.

Added new 'ui_pressed_dir' utility method for converting user input to a
vector.


assets/ui/chat/chat-choice-mood-sheet.png.import
```diff
@@ -0,0 +1,34 @@
+[remap]
```
Poobslag, line 1: Remove this unnecessary import file.

assets/ui/chat/chat-choice-mood.png.import
```diff
@@ -0,0 +1,34 @@
+[remap]
```
Poobslag, line 1: Remove this unused import file.

assets/ui/chat/chat-choice-mood0.png.import
```diff
@@ -0,0 +1,34 @@
+[remap]
```
Poobslag, line 1: Remove the rest of these unnecessary import files -- there's many of them.

assets/ui/chat/chat-theme.tres
```diff
@@ -0,0 +1,39 @@
+[gd_resource type="Theme" load_steps=6 format=2]
```
Poobslag, line 1: This theme is specifically for "chat choices" name, it appropriately

assets/ui/chat/chat-choice-mood-sheet.png.import
Poobslag: Removed.

assets/ui/chat/chat-choice-mood.png.import
Poobslag: Removed.

assets/ui/chat/chat-choice-mood0.png.import
Poobslag: Removed.

assets/ui/chat/chat-theme.tres
Poobslag: Renamed to `chat-option-theme.tres`

PR #167, opened 2020-05-05
Removed some P/Q climb kicks. Prevent initial DAS from killing you.

Poobslag:
Some P/Q climb kicks were causing the piece to jump up unexpectedly instead of
squeezing into a gap.

Added the initial spawn locations to the initial DAS kicks. It felt bad
dying as a result of applying initial DAS, so now if you die, you would
have died anyways even without initial DAS.


Poobslag:
I've marked this as a WIP in case we find more piece kicks to adjust this week. I'd like these kinds of piece kick tweaks to be bundled in one or two commits, rather than being spread out across  10-20 commits.

PR #169, opened 2020-05-06
Customer3D mesh no longer tilts. Tweaked collisions, slipping

Poobslag:
Tilted meshes could potentially be problematic because tall sprites near a wall
would tilt towards the wall and clip. This would have been a problem for
background sprites such as trees or street lights. Customer3D mesh is now
vertically stretched instead of being tilted.

Vertically stretched meshes also take up less horizontal space, so I could
shrink the hit boxes into compact cylinders.

Modified the slip logic to avoid a bug where Turbo could slip if she bumped
into a customer horizontally.

Modified the movement animation logic to ignore infinitesmally small
movement vectors. This was resulting in a glitchy flickering look when the
character pressed into a wall.

### DIFF
--- a/src/main/world/turbo.gd
+++ b/src/main/world/turbo.gd
@@ -81,13 +81,13 @@ func _physics_process(delta) -> void:
 	_apply_friction()
 	_apply_gravity(delta)
 	_apply_player_walk(delta)
-	_update_animation()
-	_update_camera_target()
 	
 	var was_on_floor := is_on_floor()
 	var old_velocity := _velocity
 	_velocity = move_and_slide(_velocity, Vector3.UP)
 	
+	_update_animation()
+	_update_camera_target()
 	_maybe_play_bonk_sound(old_velocity)
 	_maybe_slip(delta)
 	_process_jump_buffers(was_on_floor)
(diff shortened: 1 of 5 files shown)


src/main/world/turbo.gd
```diff
@@ -319,7 +322,7 @@ this logic she ends up hovering in space frequently.
 """
 func _slip_from_ledges(delta: float) -> bool:
 	if $RayCast.is_colliding():
-		# don't slip; character is sitting squarely on the ground
+		# don't slip; character is in mid-air, or squarely on the ground
```
Poobslag, line 325: The "mid-air" case is handled elsewhere; this comment is out of place

src/main/world/turbo.gd
```diff
@@ -292,20 +292,23 @@ squarely in the middle of it.
...
+		return false
+	
+	var slip_direction := Vector3.ZERO
+	var collider: Object = $RayCast.get_collider()
+	if collider.get("foothold_radius"):
+		# Turbo is standing on something she might slip from
+		slip_direction = translation - collider.translation
+		slip_direction.y = 0
+		if slip_direction.length() <= collider.foothold_radius:
+			# Turbo is too close the center, and shouldn't slip
```
Poobslag, line 306: 'too close' -> 'close'

src/main/world/turbo.gd
Poobslag: Removed mid-air comment

src/main/world/turbo.gd
Poobslag: Fixed comment

Issue #176, opened 2020-05-09
Handle case where player escapes top of overworld


Poobslag:
While the physics don't permit it, nothing prevents the objects including the player from jumping over the impassible walls of the grid. Additionally, if there were a physics glitch and the player fell through the floor or passed through a wall, they could fall forever and softlock the game.

If the player's Y coordinate goes too high (near the top of the impassible walls) or too low (below the level) they should be reset to a default coordinate.

We should also consider having some kind of "emergency reset" button they can press if they get stuck, or a safeguard so if they're in mid-air for more than 5 seconds or stuck in a wall for more than 5 seconds, they automatically get reset. Maybe this will never be an issue, but if it comes up once every 10 hours, we should still implement a fix.

Poobslag:
Closing 3D issues. Planning to replace 3D overworld with 2D overworld for GLES3 HTML5 compatibility.

Issue #175, opened 2020-05-09
Reorient/resize grass tufts to face camera


Poobslag:
Grass tufts should be 1.2x as tall, and rotated 45 degrees towards the camera, so they appear as billboard sprites in game.

This is especially difficult in Godot because its MeshLibrary export cannot export models with a custom point of origin, unless those models were imported from external 3D software [#1](https://godotforums.org/discussion/20338/is-it-possible-to-change-a-meshinstances-point-of-origin) [#2](https://github.com/godotengine/godot-proposals/issues/805).

I need to learn Blender so that I can create a polygon, apply a texture to it, rotate it 45°, modify its origin, and export it into a gltf for Godot.

Poobslag:
Additionally as we're improving the grass, there are some other tweaks we should do.

  * The grass should be sized appropriately.
  * Grass should vary its appearance. Sometimes there could be two tufts of grass behind each other, or it might just be mirrored or have a different sprite. But, it shouldn't always use the exact same sprite. At least four different versions would be nice.

PR #178, opened 2020-05-09
Overworld environment is decorated with grass, dirt, paths.

Poobslag:
The grass does not look the way I intended, I have opened #175 which I
will fix after I learn to use Blender.

The overworld environment supports sizes beyond 64x64. I'm envisioning
pacing similar to The Legend Of Zelda where the player spends about 5-10
seconds walking from one screen to the next. Individual screens should not
be larger than 64x64.

There is an invisible wall surrounding the overworld so the player can no
longer leap off into softlock land. There is no ceiling and no logic to
handle what happens if the player escapes. I've created issue #176 to capture
this.

The invisible walls are only 12 units high. They were 50 units high
initially but this caused bugs with Godot's collision detection. Specifically,
calling move_and_slide when the character was near the wall caused is_on_floor()
to permanently and erroneously return false. Shrinking the wall's height fixed
this.

There's a risk of Turbo getting stuck between the outer edge and the
impassible sprites. We should avoid placing objects too close to the outer
edge to mitigate this. A one tile gap should be enough.

Updated license/sound-license.txt to include 3D models and fonts.


src/main/environment/overworld-grid-map.gd
```diff
@@ -0,0 +1,176 @@
+tool
+class_name OverworldGridMap
+extends GridMap
```
Poobslag, line 3: Add a class-level comment

src/main/environment/overworld-grid-map.gd
```diff
@@ -0,0 +1,176 @@
...
+Repopulate the grid map with a simple world.
+
+The world contains a valley, walls on both sides, and a path down the middle.
+"""
+func regenerate_dramatic_world(regenerate: bool = true) -> void:
+	regenerate_empty_world(true, Vector2(9, 6))
+	
+	var ridge_top := 3
+	var ridge_bot := -3
+	var path := -1
```
Poobslag, line 70: Variables could have names like `ridge_top_z` to make it clear that they're y coordinates.

src/main/environment/overworld-grid-map.gd
```diff
@@ -0,0 +1,176 @@
...
+		if x > -9:
+			set_cell_item(x - 1, 0, path, TILE_PATH_CIRCLE)
+			set_cell_item(x - 1, 0, path + 1, TILE_PATH_CIRCLE)
+		path = _rand_walk(path, -2, 0)
+	
+	plant_grass()
+	autotile_paths()
+
+
+func _rand_walk(input: int, out_min: int, out_max: int) -> int:
```
Poobslag, line 104: Add a method comment.

src/main/environment/overworld-grid-map.gd
```diff
@@ -0,0 +1,176 @@
...
+	var output := input
+	var r := randf()
+	if r < 0.2 and output > out_min:
+		output -= 1
+	elif r < 0.4 and output < out_max:
+		output += 1
+	return output
+
+
+func regenerate_empty_world(regenerate: bool = true, extents: Vector2 = Vector2(16, 16)) -> void:
```
Poobslag, line 114: Add a method comment.

src/main/environment/overworld-grid-map.gd
```diff
@@ -0,0 +1,176 @@
...
+	for z in range(z_min - 1, z_max + 1):
+		set_cell_item(x_min - 1, -1, z, TILE_IMPASSABLE)
+		set_cell_item(x_max, -1, z, TILE_IMPASSABLE)
+	
+	for x in range(x_min, x_max):
+		for z in range(z_min, z_max):
+			set_cell_item(x, -1, z, TILE_BLOCK_ROCKROCK)
+
+
+func plant_grass() -> void:
```
Poobslag, line 136: Add a method comment.

src/main/environment/overworld-grid-map.gd
```diff
@@ -0,0 +1,176 @@
...
+	for cell_obj in get_used_cells():
+		var cell: Vector3 = cell_obj
+		if get_cell_item(cell.x, cell.y, cell.z) == TILE_BLOCK_ROCK \
+				and get_cell_item(cell.x, cell.y + 1, cell.z) == INVALID_CELL_ITEM:
+			# it's a bare grass tile; maybe add a grass tuft on the top
+			if randf() < 0.10:
+				set_cell_item(cell.x, cell.y + 1, cell.z, TILE_GRASS_TUFT)
+
+
+func autotile_paths() -> void:
```
Poobslag, line 146: Add a method comment.

src/main/environment/overworld-grid-map.gd
```diff
@@ -0,0 +1,176 @@
...
+				mask = Connect.set_r(mask)
+			
+			var item: int = PATH_AUTOTILE[mask][0]
+			var orientation: int = PATH_AUTOTILE[mask][1]
+			if item == TILE_PATH_CIRCLE:
+				orientation = ORIENT_UP if int(cell.x + cell.z) % 2 == 0 else ORIENT_RIGHT
+			set_cell_item(cell.x, cell.y, cell.z, item, orientation)
+
+
+func _is_path(x: int, y: int, z: int) -> bool:
```
Poobslag, line 168: Add a method comment.

src/main/environment/overworld-grid-map.gd
```diff
@@ -0,0 +1,176 @@
...
+	return get_cell_item(x, y, z) \
+			in [TILE_PATH_END, TILE_PATH_STRAIGHT, TILE_PATH_CORNER, TILE_PATH_CIRCLE]
+
+
+func _should_path_autotile_to(x: int, y: int, z: int) -> bool:
```
Poobslag, line 173: Add a method comment.

src/main/world/turbo.gd
```diff
@@ -285,17 +285,21 @@ func _jump() -> void:
 	play_movement_animation("idle")
 
 
+func over_something() -> bool:
```
Poobslag, line 288: Add a method comment.

src/main/environment/overworld-grid-map.gd
Poobslag: Added.

src/main/environment/overworld-grid-map.gd
Poobslag: Renamed to `ridge_max_z`, `ridge_min_z`, `path_z`

src/main/environment/overworld-grid-map.gd
Poobslag: Added comment.

src/main/world/turbo.gd
Poobslag: Added comment.

PR #185, opened 2020-05-12
Money is persisted, shown on overworld

Poobslag:
Split player-data, player-save


src/main/ui/ultra-button.gd
```diff
@@ -8,7 +8,7 @@ Calculates the label text for displaying a player's fastest time for ultra mode.
 """
 func get_best_text(scenario_name: String) -> String:
 	var text := ""
-	var best_result := ScenarioHistory.get_best_result(scenario_name, "seconds")
+	var best_result := PlayerData.get_best_result(scenario_name, "seconds")
```
Poobslag, line 11: In the absence of ScenarioHistory, this method should probably be called `get_best_scenario_result`

todo.txt
```diff
@@ -0,0 +1,31 @@
+*text displayed
+*icon displayed
+*truncates numbers to fit
+ *1
+ *13
```
Poobslag, line 5: remove todo

src/main/ui/ultra-button.gd
Poobslag: Renamed; also renamed `get_best_scenario_rank` and other similar methods

todo.txt

PR #200, opened 2020-05-15
New grade scale; eliminated D and C ranks.

Poobslag:
This new scale follows the progression of our eventual star unlocks.
'B' and 'A' grades will eventually be displayed as dots and triangles. These
will be displayed to the user and help them gauge their progress. Unbeknownst
to them at the start of the game, they have no impact on progress.

S grades will eventually be displayed as stars, like the stars of a
restaurant. Stars will be a prerequisite to hire new employees, grow your
restaurant and increase your income. All of these steps are optional but they
help to progress through the game faster.

These grades

### DIFF
--- a/src/main/global.gd
+++ b/src/main/global.gd
@@ -38,32 +38,29 @@ func _process(delta: float) -> void:
 
 
 """
-Converts a numeric grade such as '12.6' into a grade string such as 'A+'.
+Converts a numeric grade such as '12.6' into a grade string such as 'S+'.
 """
 func grade(rank: float) -> String:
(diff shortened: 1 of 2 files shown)


src/main/global.gd
```diff
@@ -42,28 +42,25 @@ Converts a numeric grade such as '12.6' into a grade string such as 'A+'.
 """
 func grade(rank: float) -> String:
```
Poobslag, line 43: method comment is inaccurate now

Poobslag: Fixed method comment

PR #206, opened 2020-05-16
Fixed several rewind-related chat bugs.

Poobslag:
There was a crash in chat-choices.gd:142 caused because the objects in an
array would change types randomly; sometimes they would be
'chat-choice-label.gd' objects and sometimes they would be
'chat-choice-button.gd' objects.

There was a bug where rewinding/fast-forwarding around chat choices could
cause errant behavior. Sometimes the rest of a conversation could be
skipped. Sometimes you'd be prompted the same question over and over.

Extracted chat-advancer.gd from chat-ui.gd. This class encapsulates the
rewind and fast-forward functionality, as well as the chat history. It's a
good candidate for unit testing, although we might not need unit tests
unless this class causes further problems.


Poobslag:
Here are the steps for reproducing the bugs I fixed:

  *`chat-choices.gd:142 'Trying to assign value of type 'chat-choice-label.gd' to a variable of type 'chat-choice-button.gd'` -- This happened when fast-forwarding/rewinding frantically, back-and-forth.
  * Infinite prompt: This happened when you're prompted -> rewind -> advance -> choose. After advancing and choosing, you're prompted again and again.
  * Dialog skipped: This happened when you choose -> rewind -> advance. Advancing causes the remaining dialog to be skipped.

todo.txt
```diff
@@ -0,0 +1,22 @@
...
+ *position = nice:0
+ *events[nice][0] has no links; stays put
+*repeat dialog in history
+ *make choice, rewind, forward, rewind, forward... duplicate dialog in history
+*dialog skipped?
+ *make choice, rewind, advance... skipped?
+*refactor
+ *???
+ *'if chat_polish.decrement_line' could be cleaned up?
+ *runs in game; no crashes
```
Poobslag, line 13: Remove this todo.txt

PR #221, opened 2020-05-20
Merged scenario huds into a single hud

Poobslag:
Scenario had a separate ScoreHud, TimeHud, LinesHud and CustomerHud.
They've all been merged into one.

The new huds use a generic white-on-cyan progress bar. This was the result
of a typo where instead of an 'rgb' color, the progress bar was
initialized with an 'rgg' color, making it cyan instead of green. This
progress bar isn't permanent anyways.

Extracted format_duration into StringUtils.

Extracted Milestone from ScenarioSettings.


src/main/puzzle/scenario.gd
```diff
@@ -21,91 +25,123 @@ func _ready() -> void:
...
+				$Hud/ProgressBar.max_value = Global.scenario_settings.level_ups[_level + 1].value
 			else:
 				# final level of marathon mode; fill up the bar as they near their goal
-				$LinesHud/ProgressBar.min_value = Global.scenario_settings.level_ups[new_level].value
-				$LinesHud/ProgressBar.max_value = winish_condition.value
+				$Hud/ProgressBar.min_value = Global.scenario_settings.level_ups[_level].value
+				$Hud/ProgressBar.max_value = winish_condition.value
+
+
+func _winish_type() -> int:
```
Poobslag, line 143: This seems like a useful utility method but it's never referenced

src/main/puzzle/scenario.gd
```diff
@@ -120,60 +156,58 @@ func _check_for_match_end() -> void:
...
+		Milestone.SCORE:
+			$Hud/Label.text = "Time"
+			$Hud/Value.set("custom_fonts/font", _xolonium_36)
+			$Hud/ProgressBar.max_value = winish_condition.value
+			$Hud/ProgressBar.value = 0
+	
+	_update_hud_colors()
+	if winish_condition.type == Milestone.LINES:
+		# update level display
+		_update_hud_content()
```
Poobslag, line 210: This if statement seems out-of-place after the preceding match statement

src/main/puzzle/scenario.gd
Poobslag: Code now references _winish_type() utility method.

src/main/puzzle/scenario.gd
Poobslag: Pulled _update_hud_content out of the if statement.

PR #222, opened 2020-05-21
Moved MilestoneHud into Puzzle. Added u-kicks.

Poobslag:
Started introducing tutorial UI element customization. Launching the
tutorial scenario hides the chalkboard, summons master fat.

Simplified milestone hud logic. Instead of granular methods like 'update
colors', 'update content', 'update goal', 'prepare', there's just two methods
now: 'prepare hud' and 'update hud'

Added some missing u-kick tests. The piece was stuck even though there were
clear places it could rotate.

### DIFF
--- a/src/main/global.gd
+++ b/src/main/global.gd
@@ -20,6 +20,12 @@ var scenario_settings := ScenarioSettings.new()
 # The customers who will show up during the next puzzle. The first customer in the queue will show up first.
 var customer_queue := []
 
+# 'true' if the customer should change when breaking combo. Disabled for tutorials.
+var customer_switch := true
+
+# 'true' if the customer should get fatter. Disabled for tutorials.
+var customer_fatten := true
+
 # 'true' if launching a puzzle from the overworld. This changes the menus and disallows restarting.
 var overworld_puzzle := false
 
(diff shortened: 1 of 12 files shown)


src/main/global.gd
```diff
@@ -19,6 +19,8 @@ var scenario_settings := ScenarioSettings.new()
 
 # The customers who will show up during the next puzzle. The first customer in the queue will show up first.
 var customer_queue := []
+var customer_switch := true
+var customer_fatten := true
```
Poobslag, line 23: Document these properties.

todo.txt
```diff
@@ -0,0 +1,92 @@
...
+  -shows message
+ -prepares playfield
+ -only o pieces
+ -communication before
+ -communication after
+ -deletes your last piece if you didn't do a squish move
+ -progress after doing two squish moves
+-super ultra skip
+ -double score for cake piece in part 1
+ -OH, MY!!! (very big font)
```
Poobslag, line 82: Remove todo

src/main/global.gd
Poobslag: Added comments.

todo.txt

PR #227, opened 2020-05-22
Further fleshed out level editor; save/load, drag/drop

Poobslag:
Level editor now supports saving/loading. Levels can be loaded from a
resource or from an external file.

Can drag/drop vegetables to playfield in scenario editor. These blocks are
saved in the json model and appear when starting a scenario.

Extracted PuzzleTileMap, PuzzleTileSet from Playfield, Puzzle. This tilemap was
needed for the level editor, and it didn't make sense to define the same
tileset/tilemap in three places.

Added json beautifier add-on. This allows our generated scenarios to span
multiple lines instead of being crammed together in a single row.

Added license info for Gut, json beautifier


todo.txt
```diff
@@ -0,0 +1,97 @@
+-scenario editor
+ -edit layout
+  *visual stuff
```
Poobslag, line 3: Remove this todo file.

PR #228, opened 2020-05-22
Level editor lets you add pieces to the playfield.

Poobslag:
Resolved 'tile_map.has(p_pid)' errors in PlayfieldFx.

Refactored many places to use 'PuzzleTileMap.set_block' utility method instead
of 'TileMap.set_cell'


src/main/puzzle/editor/piece-chunk.gd
```diff
@@ -0,0 +1,31 @@
...
+	EditorPiece.PIECE_P: PieceTypes.piece_p,
+	EditorPiece.PIECE_Q: PieceTypes.piece_q,
+	EditorPiece.PIECE_T: PieceTypes.piece_t,
+	EditorPiece.PIECE_U: PieceTypes.piece_u,
+	EditorPiece.PIECE_V: PieceTypes.piece_v,
+}
+
+var _orientation := 0
+
+func _refresh_tilemap() -> void:
```
Poobslag, line 26: You actually got the code order right on the first try, for these 7 different components without looking it up? Wow, that's hard mode...

PR #231, opened 2020-05-23
Levels accept piece types, score variations. Piece queue changes

Poobslag:
Changed piece queue to always start with five bad pieces

The old approach of 'two boxes at the start' was kind, and maybe i'll
bring it back. I'm trying something different for now.

### DIFF
--- a/src/main/puzzle/puzzle.gd
+++ b/src/main/puzzle/puzzle.gd
@@ -87,6 +87,18 @@ func milevalue() -> Label:
 	return $Chalkboard/MilestoneHud/Value as Label
 
 
+func set_block(pos: Vector2, tile: int, autotile_coord: Vector2 = Vector2.ZERO) -> void:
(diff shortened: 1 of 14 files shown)


src/main/puzzle/puzzle.gd
```diff
@@ -87,6 +87,17 @@ func milevalue() -> Label:
...
+	$Playfield.set_block(pos, tile, autotile_coord)
+
+
+func set_piece_types(types: Array) -> void:
+	$NextPieceDisplays.set_piece_types(types)
+
+
+func set_piece_start_types(types: Array) -> void:
+	$NextPieceDisplays.set_piece_start_types(types)
+
```
Poobslag, line 100: Extra line of whitespace needed

Poobslag: Added

PR #232, opened 2020-05-23
Player can survive top out. Refactored scenario rules.

Poobslag:
Scenario json rules are now formatted as 'top-out-1' instead of 'top-out
1'. The new format is less ambiguous and easier to parse.

New scenario settings: top-out-penalty, lose-condition. Marathon is set to
'top-out 1' so the player loses if they top out. The default is 'top-out
3' which gives players some leeway, while still mercy killing a scenario
if it's too hard for them.

Moved cake/snack constants and utility methods from Playfield to
PuzzleTileMap.

Fixed bug where marathon started out with bar half-full

Extracted rules classes from ScenarioSettings, consolidated rule parsing
logic into RuleParser. I expect to iterate on RuleParser's design as I
create more rules and figure out what they have in common.


src/main/player-data.gd
```diff
@@ -51,7 +51,7 @@ func get_best_scenario_result(scenario: String, property: String = "score") -> R
 	if scenario_history.has(scenario):
 		for rank_result in scenario_history[scenario]:
 			if property == "seconds":
-				if rank_result.died:
+				if rank_result.topped_out():
```
Poobslag, line 54: Losing should disqualify your score. Topping out is acceptable

PR #248, opened 2020-05-25
Data load now handles missing fields. Renamed save file.

Poobslag:
Json data is now loaded using json.get("foo", "default") instead of
json["foo"]. This ensures that old save data is more forwards-compatible,
and people won't need to start over every time the save format is changed
slightly.

In an additional attempt to reduce the number of save-breaking changes, save
data is now stored in turbofat0.save. When additional save slots are added, they
can be named turbofat1.save and turbofat2.save

Json methods now have universal names like 'from_json_dict',
'from_json_string_array', 'to_json_dict'

### DIFF
--- a/src/main/puzzle/scenario/blocks-start.gd
+++ b/src/main/puzzle/scenario/blocks-start.gd
@@ -7,18 +7,16 @@ var used_cells := []
 var tiles := {}
 var autotile_coords := {}
 
-"""
-Populates this object from a json dictionary.
-"""
-func from_dict(dict: Dictionary) -> void:
-	if dict.has("tiles"):
-		for json_tile in dict["tiles"]:
-			var json_pos_arr: PoolStringArray = json_tile["pos"].split(" ")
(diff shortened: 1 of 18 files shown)


src/main/puzzle/scenario/blocks-start.gd
```diff
@@ -7,18 +7,16 @@ var used_cells := []
...
-	if dict.has("tiles"):
-		for json_tile in dict["tiles"]:
-			var json_pos_arr: PoolStringArray = json_tile["pos"].split(" ")
-			var json_tile_arr: PoolStringArray = json_tile["tile"].split(" ")
-			var pos := Vector2(int(json_pos_arr[0]), int(json_pos_arr[1]))
-			var tile := int(json_tile_arr[0])
-			var autotile_coord := Vector2(int(json_tile_arr[1]), int(json_tile_arr[2]))
-			set_block(pos, tile, autotile_coord)
+func from_json_dict(json: Dictionary) -> void:
+	for json_tile in json.get("tiles", []):
```
Poobslag, line 11: This default value might be redundant.

PR #253, opened 2020-05-25
Slower lock times for novices. Faster DAS for experts.

Poobslag:
A playtester reported that the lock time of 40 frames felt too oppressive
for what's meant to be a casual game. I've slowed it down to 60 frames for
the beginner levels (0-9)

I found the DAS reduction from FA to FB to be jarring when progressing
through expert difficulties; the pieces felt out of control with only 5 frames
of DAS. I've decreased it to 5 frames for FA as well, so now it feels out of
control all the time. Ha ha.

We can let players adjust their own DAS if they hate it. 5 frames feels OK
to me now that I've practiced it.

### DIFF
--- a/src/main/puzzle/piece/piece-speeds.gd
+++ b/src/main/puzzle/piece/piece-speeds.gd
@@ -28,19 +28,19 @@ var _speeds := {}
 
 func _ready() -> void:
 	# tutorial; piece does not drop
-	_add_speed(PieceSpeed.new("T",   0, 20, 36, 7, 16, 40, 24, 12))
+	_add_speed(PieceSpeed.new("T",   0, 20, 36, 7, 16, 60, 24, 12))
 	
 	# beginner; 10-30 blocks per minute
-	_add_speed(PieceSpeed.new("0",   4, 20, 36, 7, 16, 40, 24, 12))
-	_add_speed(PieceSpeed.new("1",   5, 20, 36, 7, 16, 40, 24, 12))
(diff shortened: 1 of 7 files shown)


src/main/puzzle/piece/piece-speeds.gd
```diff
@@ -63,9 +63,9 @@ func _ready() -> void:
 	_add_speed(PieceSpeed.new("AF", 20*G, 16, 12, 7, 10, 30, 12,  6))
 	
 	# crazy; 120-250 blocks per minute
-	_add_speed(PieceSpeed.new( "F0",    4, 12, 8, 7, 10, 24, 6, 3))
-	_add_speed(PieceSpeed.new( "F1",  1*G, 12, 8, 7, 10, 24, 6, 3))
-	_add_speed(PieceSpeed.new( "FA", 20*G, 12, 8, 7, 10, 24, 6, 3))
+	_add_speed(PieceSpeed.new( "F0",    4, 12, 8, 5, 10, 24, 6, 3))
+	_add_speed(PieceSpeed.new( "F1",  1*G, 12, 8, 5, 10, 24, 6, 3))
+	_add_speed(PieceSpeed.new( "FA", 20*G, 12, 8, 5, 10, 24, 6, 3))
```
Poobslag, line 68: What are these three '10' numbers near the end that didn't change? Is it related to DAS?

Poobslag: Those '10 numbers were the DAS; I changed the prelock by mistake.

I've still reduced the prelock for consistency, but changed the DAS from 10/7 to an average 8.

Issue #234, opened 2020-05-24
All exported json should include version flag


Poobslag:
Save data is currently very in flux; last month the save data just included a list of histories. This month we are storing two kinds of data, history data and the player's money. Next month it's likely we'll change the scenario history to only include 2-3 records per scenario. This will also be a compatibility-breaking changes.

We only have 2 users right now, so this change is frustrating. When we have 10 users and some of them have put hours into progressing through the game, this will be a game-ruining issue. We need to be able to tell when we're working with an old save format, and we need to have utilities to convert or load old data.

Exported scenarios and save data should include a version flag, like `"version": "15c0"`. The first digit corresponds to the (year - 2019) in hex. The second digit corresponds to the month (1-12) in hex. The next two digits correspond to the (day  * 8) in hex, plus a little extra.

For 2020-05-24, this would be in the range `15c0`-`15c7`.

The version should only be incremented if the save format changes.

Poobslag:
In OpenOffice, this format is expressed with `=DEC2HEX((YEAR(NOW())-2019)*4096+(MONTH(NOW()))*256+(DAY(NOW()))*8+INT(HOUR(NOW()) / 3))`

PR #256, opened 2020-05-26
Saved data is pure json, includes version numbers.

Poobslag:
Scenario, player save, dialog data includes version numbers. The version
number follows a 'ymdh' format which is documented in issue #234.

Fixed failing tests related to recent piece speed changes.


src/main/player-save.gd
```diff
@@ -9,6 +9,63 @@ Save data is stored as a series of lines. Each line contains a 4-character strin
 	'scen{...}': Scenario data; how many points they scored, how long they took
```
Poobslag, line 9: These comments are inaccurate.

Poobslag: Removed comments; the new format is less surprising than the old format, and does not warrant documentation.

PR #258, opened 2020-05-27
Save data is backwards-compatible.

Poobslag:
On the surface it's a waste of time to spend 2-3 hours writing code to
maintain save compatibility, when there is literally a single user.
However I think this will continue to be a problem in future versions of
the game, so it's important to establish a backwards compatibility
framework early.

Removed JSON formatting for player save data. This caused a noticable
stutter (~300 ms) when clearing a level.

### DIFF
--- a/src/main/global.gd
+++ b/src/main/global.gd
@@ -194,3 +194,10 @@ static func get_file_as_text(path: String) -> String:
 	var text := f.get_as_text()
 	f.close()
 	return text
+
+
+static func write_file(path: String, text: String) -> void:
+	var f := File.new()
+	f.open(path, f.WRITE)
+	f.store_string(text)
+	f.close()
--- a/src/main/player-save.gd
+++ b/src/main/player-save.gd
@@ -62,8 +62,11 @@ class SaveItem:
 # This version number follows a 'ymdh' hex date format which is documented in issue #234.
 const PLAYER_DATA_VERSION := "15d2"
 
-# filename to use when saving/loading player data. can be changed for tests
+# Filename to use when saving/loading player data. Can be changed for tests
 var player_data_filename := "user://turbofat0.save"
+
+# Provides backwards compatibility with older save formats
+var old_save := OldSave.new()
 
 func _ready() -> void:
 	load_player_data()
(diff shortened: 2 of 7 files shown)


src/main/player-save.gd
```diff
@@ -81,28 +83,27 @@ func save_player_data() -> void:
 		for rank_result in PlayerData.scenario_history[key]:
 			rank_results_json.append(rank_result.to_json_dict())
 		save_json.append(SaveItem.named("scenario-history", key, rank_results_json).to_json_dict())
-	
-	var save_game := File.new()
-	save_game.open(player_data_filename, File.WRITE)
-	save_game.store_string(JSONBeautifier.beautify_json(to_json(save_json), 1))
-	save_game.close()
+	Global.write_file(player_data_filename, to_json(save_json))
+	pass
```
Poobslag, line 87: Unnecessary `pass` statement

src/main/player-save.gd
```diff
@@ -65,6 +65,8 @@ const PLAYER_DATA_VERSION := "15d2"
 # filename to use when saving/loading player data. can be changed for tests
 var player_data_filename := "user://turbofat0.save"
 
+var old_save := OldSave.new()
```
Poobslag, line 68: Document this field, just to make it clear it's for backwards compatibility, and not for a player's previous save file

src/main/global.gd
```diff
@@ -194,3 +194,10 @@ static func get_file_as_text(path: String) -> String:
 	var text := f.get_as_text()
 	f.close()
 	return text
+
+
+static func write_file(path: String, text: String) -> void:
+	var f := File.new()
+	f.open(path, f.WRITE)
+	f.store_string(text)
+	f.close()
```
Poobslag, line 203: We're starting to have a few of these kinds of functions. We should consider splitting out `utils` for utility methods, and `global` for global properties. Something for next week.

src/main/player-save.gd
Poobslag: Documented, `Provides backwards compatibility with older save formats`

src/main/player-save.gd
Poobslag: Removed

PR #267, opened 2020-05-30
Designed new practice menu with high scores.

Poobslag:
'Marathon' is now 'Survival'. This is the only mode where topping out ends
your game, so 'survival' is a more fitting name.

Fixed bug where first 5 pieces could include two 'O' or 'T' pieces.


assets/puzzle/scenario/survival-master.json
```diff
@@ -9,29 +9,29 @@
...
+      "value": "300",
       "level": "FF"
     }
   ],
   "win-condition": {
     "type": "lines",
-    "value": "1000",
-    "lenient-value": "500"
+    "value": "500",
+    "lenient-value": "300"
```
Poobslag, line 34: Worth documenting in commit comment.

src/main/player-data.gd
```diff
@@ -24,44 +24,52 @@ var history_size := 1000
 var _rank_calculator := RankCalculator.new()
 
 """
-Calculates the best rank a player's received for a specific scenario.
+Calculates the best a player's best performances for a specific scenario.
```
Poobslag, line 27: Strange grammar; should be `Calculates a player's best performances`

assets/puzzle/scenario/survival-master.json
Poobslag: Amended commit comment, `Reduced marathon survival from 1,000 to 500 lines`

src/main/player-data.gd
Poobslag: Fixed grammar, `Returns a player's best performances...`

PR #271, opened 2020-05-30
Player can customize volume with new settings menu.

Poobslag:
The settings menu can be displayed by pressing the 'Esc' key or clicking
the settings button. The volume settings are saved alongside their money
and rank history.

Volume settings default to 50%, a little higher than what they used to be
(44%)

Extracted test-backwards-compatible.gd from test-player-data.gd


src/test/test-backwards-compatible.gd
```diff
@@ -0,0 +1,72 @@
...
+
+func after_each() -> void:
+	PlayerData.history_size = 1000
+	var save_dir := Directory.new()
+	save_dir.open("user://")
+	save_dir.remove(TEMP_FILENAME)
+	save_dir.remove(TEMP_FILENAME_0517)
+
+
+func test_backwards_compatible_lost() -> void:
```
Poobslag, line 27: Now that they're in their own class, these methods can have less repetitive names

Poobslag: Changed to `test_lost_true`, `test_lost_false`, etc

PR #273, opened 2020-05-30
Scenario history is saved more efficiently.

Poobslag:
The previous design which saved the newest 1,000 (!) scores for each mode had
two flaws.

1. Each score takes about 500 bytes of disk space, so 1,000 scores would
take up 500k of memory and disk space for a single scenario.

2. In the unlikely change that a player played a single scenario more than
1,000 times, the previous strategy would always kick out the oldest record
which could be the player's high score. It would take a great player 5-6 hours
of playing nothing but Ultra Normal to finally kick out their oldest
score, so this is unlikely but it could happen.

The new design saves a maximum of 7 scores: the player's newest score,
their top 3 daily scores, and the top 3 all-time scores.

Extracted ScenarioHistory from PlayerData.

### DIFF
--- a/src/main/puzzle/scenario/scenario-history.gd
+++ b/src/main/puzzle/scenario/scenario-history.gd
@@ -1,0 +1,118 @@
+class_name ScenarioHistory
+"""
+Stores the best results the player has achieved for different scenarios. This is currently used for calculating high
+scores, but could eventually be used for displaying statistics or calculating rewards.
+"""
+
+# how many daily and all-time records we can store before we start deleting old ones
+var max_size := 3
+
+# key: scenario name
+# value: array of RankResults for the specified scenario
+var rank_results := {}
+
+func scenario_names() -> Array:
+	return rank_results.keys()
+
+
+"""
+Resets the scenario history to its default empty state.
+"""
+func reset() -> void:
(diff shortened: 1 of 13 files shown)


src/main/puzzle/scenario/scenario-history.gd
```diff
@@ -0,0 +1,116 @@
...
+		results.sort_custom(self, "_compare_by_seconds")
+	else:
+		results.sort_custom(self, "_compare_by_score")
+	return results.slice(0, max_size - 1)
+
+
+func prev_result(scenario: String) -> RankResult:
+	if not rank_results.has(scenario) or rank_results[scenario].empty():
+		return null
+	return rank_results[scenario][0]
```
Poobslag, line 70: This method assumes rank_results is ordered by date, but prune() shuffles the array.

Poobslag: Surprisingly, prune() was preserving the array's order. Whatever algorithm gdscript's dictionaries use when adding new keys preserves the order in which keys were inserted.

I still changed the algorithm, just in case this behavior changes in the future. It would result in some very confusing and inconsistent bugs.

PR #274, opened 2020-05-31
All puzzle modes display distance to goal.

Poobslag:
A bug had been introduced where ultra no longer displayed something like
'$1000' on startup, and only displayed '2:30' in a tiny font. Upon further
consideration, I realized all modes for consistency should just display
the progress towards the goal.

Survival now displays the lines remaining. Ultra displays score remaining.
Sprint displays time remaining. The '5 customers' mode displays customers
remaining.

Also refactored the milestone label to use the same 'FontFitLabel' logic
as ChatChoiceLabel.

Extracted milestone-hud script from scenario.gd.

PuzzleScore now maintains the current level, and provides signals when
lines are cleared or the level is incremented.

Removed secret 'DB' level for easter egg. Level is no longer displayed,
and with the milestone hud changes this easter egg was becoming
cumbersome.

Made SettingsMenu components invisible so they don't block scenes in the
editor.

### DIFF
--- a/src/main/puzzle/puzzle-score.gd
+++ b/src/main/puzzle/puzzle-score.gd
@@ -14,28 +14,38 @@ signal game_prepared
 # emitted after everything's been erased in preparation for a new game.
 signal after_game_prepared
 
+# emitted after the scenario has customized the puzzle's settings.
+signal after_scenario_prepared
+
 signal game_started
 signal game_ended
-signal score_changed(value)
-signal bonus_score_changed(value)
+signal level_index_changed(value)
+
+# These four signals are always emitted in this order: customer_score, bonus_score, score, lines
 signal customer_score_changed(value)
+signal bonus_score_changed(value)
+signal score_changed(value)
+signal lines_changed(value)
+
 signal combo_ended
 
 # Player's performance on the current scenario
 var scenario_performance := ScenarioPerformance.new()
 
+# The scores for each customer in the current scenario (bonuses and line clears)
+var customer_scores := [0]
+
 # Bonus points awarded during the current combo. This only includes bonuses
 # and should be a round number like +55 or +130 for visual aesthetics.
 var bonus_score := 0
-
-# The scores for each customer in the current scenario (bonuses and line clears)
-var customer_scores := [0]
 
 # 'true' if the player has started a game which is still running.
 var game_active: bool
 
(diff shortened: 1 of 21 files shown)


src/main/puzzle/puzzle-score.gd
```diff
@@ -52,14 +62,33 @@ func prepare_game() -> void:
...
-
-	bonus_score = 0
-	emit_signal("bonus_score_changed", 0)
+	emit_signal("score_changed", scenario_performance.score)
+	emit_signal("lines_changed", scenario_performance.lines)
 
+	set_bonus_score(0)
+	
 	customer_scores = [0]
 	emit_signal("game_prepared")
```
Poobslag, line 71: This signal is emitted out of order (customer score, bonus score, score, lines)

src/main/puzzle/puzzle-score.gd
```diff
@@ -92,47 +121,49 @@ func add_line_score(combo_score: int, box_score: int) -> void:
...
 	
-	scenario_performance.lines += 1
 	scenario_performance.combo_score += combo_score
 	scenario_performance.box_score += box_score
-	
-	_add_score(1)
 	_add_bonus_score(combo_score + box_score)
+	_add_score(1)
+	_add_line()
 	_add_customer_score(1 + combo_score + box_score)
```
Poobslag, line 129: Out of order; customer score should be updated first

src/main/puzzle/puzzle-score.gd
```diff
@@ -14,29 +14,39 @@ signal game_prepared
 # emitted after everything's been erased in preparation for a new game.
 signal after_game_prepared
 
+# the scenario has customized all of the settings it needs to.
```
Poobslag, line 17: sentence fragment

src/main/puzzle/puzzle-score.gd
Poobslag: Signals are now emitted back-to-back after all values are assigned.

src/main/puzzle/puzzle-score.gd
Poobslag: Signals are now emitted back-to-back in an appropriate order

src/main/puzzle/puzzle-score.gd
Poobslag: Fixed fragment

PR #276, opened 2020-05-31
High score system no longer relies on 'ultra-' prefix

Poobslag:
The old logic looked at a scenario name to see if it started with
'ultra-', and those scenarios were compared differently. This is a poor
design because it requires all goal-based scenarios to start with the word
'ultra-'.

Now, we take any scenarios with a 'score' goal and give them a different
'compare' property. For now this can be '+score' or '-seconds', but in the
future other scenarios could have other definitions of what it means to
'do a good job at a level'. Some might compare how many or how few customers
you served, reaching a target score in as few lines as possible, or
maximizing the number of boxes made. These could all be represented in
data as something like '+customers', '-customers', '-lines' or
'+box_score'.

### DIFF
--- a/src/test/test-backwards-compatible.gd
+++ b/src/test/test-backwards-compatible.gd
@@ -65,6 +65,22 @@ func test_timestamp_created() -> void:
 	assert_true(history_ultra.timestamp.has("second"))
 
 
+func test_compare_seconds() -> void:
+	assert_true(PlayerData.scenario_history.scenario_names().has("ultra-normal"))
+	var history_ultra: RankResult = PlayerData.scenario_history.results("ultra-normal")[0]
+	
+	# ultra records should be compared by lowest seconds. this 'compare' field didn't exist before
+	assert_eq(history_ultra.compare, "-seconds")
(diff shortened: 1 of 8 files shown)


src/test/test-backwards-compatible.gd
```diff
@@ -65,6 +65,14 @@ func test_timestamp_created() -> void:
 	assert_true(history_ultra.timestamp.has("second"))
 
 
+func test_ultra_compare() -> void:
+	assert_true(PlayerData.scenario_history.scenario_names().has("ultra-normal"))
+	var history_ultra: RankResult = PlayerData.scenario_history.results("ultra-normal")[0]
+	
+	# new rank results include a 'compare' field so we know how to compare them
+	assert_eq(history_ultra.compare, "-seconds")
```
Poobslag, line 73: Add a counterpart to this test to ensure 'sprint' records have the correct 'compare' value

Poobslag: Added `test_compare_score` test.

PR #286, opened 2020-06-02
Massively reorganized assets and source.

Poobslag:
Assets are now in /assets/main, /assets/test, /assets/demo

Moved .tres assets from /assets into /src. GDQuest suggests /assets
should have shared files like images, sounds, music and text produced
outside Godot, while /src should have all scenes and GDScript files. .tres
files are generated by Godot, so they seem like they belong in the /src
directory.

Renamed 'customer' to 'creature'. 'Customer' is an unsuitable name since
it needs to apply to Spira and Turbo who are not customers.


Poobslag:
...I'm not reviewing this. 939 is a big number. I hope you knew what you were doing.

PR #287, opened 2020-06-02
Added trailing commas to arrays, dictionaries

Poobslag:


Poobslag:
Those commas look comma-riffic. That was some time well spent. Make sure to update the guidelines in the wiki, too.

PR #288, opened 2020-06-02
Pushed Scenario logic into child classes.

Poobslag:
The Scenario class previously had logic for resetting the score, initializing the
playfield, loading up the little 'tutorial guy', saving the results, and
other stuff. This has all been pushed into different places. Instead,
scenario.gd manages loading/saving of scenario files (previously handled
by ScenarioLibrary)

PuzzleScore now handles the 'puzzle lifecycle' which was previously
managed by puzzle. It prepares a puzzle, pauses and starts the puzzle. It
calculates when the player wins or loses and emits appropriate signals.

Global.scenario_settings is now Scenario.settings. A new
Scenario.settings_changed signal gives classes a logical place to
initialize themselves if they have scenario-dependent logic, such as the
PieceQueue which needs to change its pieces if the scenario uses a
nonstandard piece set.

Moved puzzle-related scripts and assets from /puzzle/scenario into
/puzzle. /puzzle/scenario now only includes scripts and assets related to
saving, loading, and maintaining scenario settings and scenario history.

Moved tutorial-related scripts and assets from /puzzle into
/puzzle/tutorial.


src/main/global.gd
```diff
@@ -11,9 +11,6 @@ const NUM_SCANCODES: Dictionary = {
 # Target number of creature greetings (hello, goodbye) per minute
 const GREETINGS_PER_MINUTE := 3.0
 
-# The scenario currently being launched or played
-var scenario_settings := ScenarioSettings.new()
-
 # The scenario which was initially launched. Some tutorial scenarios transition
 # into other scenarios, so this keeps track of the original.
 var launched_scenario_name
```
Poobslag, line 16: This should probably stay paired with `scenario_settings` since they're so closely related

src/main/puzzle/puzzle.gd
```diff
@@ -5,19 +5,23 @@ Represents a minimal puzzle game with a piece, playfield of pieces, and next pie
 class to add goals, win conditions, challenges or time limits.
 """
 
-# emitted a few seconds after the game ends, for displaying messages
-signal after_game_ended
-
 signal topped_out
 
-onready var _go_voices := [$GoVoice0, $GoVoice1, $GoVoice2]
```
Poobslag, line 13: This new 'PuzzleSfx' class might warrant a commit comment.

src/main/puzzle/puzzle.gd
```diff
@@ -52,78 +56,14 @@ func show_detail_message(text: String) -> void:
...
-	
-	sound.play()
-	PuzzleScore.end_game()
-	show_message(message)
-	yield(get_tree().create_timer(delay), "timeout")
-	emit_signal("after_game_ended")
-
-
 func get_milestone_hud() -> MilestoneHud:
 	return $Chalkboard/MilestoneHud as MilestoneHud
```
Poobslag, line 60: Give these methods like get_milestone_hud() a look to make sure they're still used. Many of them like `set_chalkboard_visible` were removed already but we may have overlooked some.

src/main/global.gd
Poobslag: Moved `overworld_puzzle` and `launched_scenario_name` into Scenario.

src/main/puzzle/puzzle.gd
Poobslag: Added comment, `Extracted start-end-sfx.gd from Puzzle class.`

src/main/puzzle/puzzle.gd
Poobslag: Removed several unused delegate methods; `PuzzleHud.show_detail_message()`, `Puzzle.get_next_piece_displays()`, `Puzzle.show_detail_message()`, `Puzzle.get_milestone_hud()`, `Puzzle.set_block()`

PR #291, opened 2020-06-03
Seat now references a customer, doesn't 'own' a customer

Poobslag:


src/main/world/restaurant/creature-loader.gd
```diff
@@ -51,6 +51,10 @@ func load_details(creature_def: Dictionary) -> void:
 	_load_colors(creature_def)
 
 
+static func random_def() -> Dictionary:
```
Poobslag, line 54: Document this function

src/main/world/restaurant/seat.gd
```diff
@@ -6,59 +6,33 @@ contain the shadows itself, since they need to appear behind other objects outsi
...
+	_creature = creature
+	refresh()
 
 
-"""
-Plays a 'check please!' voice sample, for when a creature is ready to leave
-"""
-func play_goodbye_voice() -> void:
-	$Creature.play_goodbye_voice()
+func refresh() -> void:
```
Poobslag, line 17: Should this be private? Either way, comment it

Poobslag: It's referenced outside the class, so it can't be private. I've added a comment.

src/main/world/restaurant/creature-loader.gd
Poobslag: Added comment

PR #312, opened 2020-06-06
North facing mouth frame '2'. Fixed creature behavior in editor.

Poobslag:
Mouth sprites now have north-facing mouth in frame 2. This simplifies our logic
for facing towards/away if our animation tools know which frame to switch to
when a creature changes direction. Otherwise each mouth would need to set this
property manually.

Fixed creature behavior in editor. The creature used to be viewable in the
editor, but this was disrupted by recent refactorings.

Removed 'tool' declaration from OverworldGridMap, I suspect this declaration is
why the scene was constantly showing up with huge changes in Git.

Removed 'tool' declaration from skill tally item, I can't remember why this was
needed.


src/main/world/creature/Creature.tscn
```diff
@@ -264,7 +264,7 @@ tracks/0/keys = {
 "times": PoolRealArray( 0, 0.0333333, 0.0666667, 0.1, 0.133333, 0.166667, 0.2, 0.233333, 0.266667, 0.3, 0.333333, 0.366667, 0.4, 0.433333, 0.599999 ),
 "transitions": PoolRealArray( 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1 ),
 "update": 1,
-"values": [ 1, 2, 3, 4, 5, 6, 7, 1, 7, 8, 8, 7, 7, 1, 1 ]
+"values": [ 1, 3, 4, 5, 6, 7, 8, 1, 8, 7, 7, 6, 6, 1, 1 ]
```
Poobslag, line 267: This doesn't quite look right, shouldn't it have changed to `8, 9, 9, 8, 8`?

Poobslag: Changed to ` [ 4, 5, 6, 7, 8, 1, 8, 9, 9, 8, 8, 1, 1 ]`

PR #320, opened 2020-06-07
Added gamepad support.

Poobslag:
InputMappings now use 'escape' key as both ui_cancel and ui_menu. This
requires some unusual logic to handle the case where both actions are
activated at once.

Disabled 'shortcut in tooltip' for buttons.

Due to Godot #28914, InputEventAction shortcuts have crazy tooltips like
(InputEventAction : action=test, pressed=(true))

### DIFF
--- a/src/main/ui/menu/settings-menu.gd
+++ b/src/main/ui/menu/settings-menu.gd
@@ -10,6 +10,9 @@ signal quit_pressed
 # The text on the menu's quit button
 export (String, "Quit", "Save + Quit") var quit_text: String setget set_quit_text
 
+# the UI control which was focused before this settings menu popped up
+var _old_focus_owner: Control
+
 func _ready() -> void:
 	# starts invisible
 	hide()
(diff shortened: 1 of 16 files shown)


src/main/ui/menu/settings-menu.gd
```diff
@@ -10,6 +10,8 @@ signal quit_pressed
 # The text on the menu's quit button
 export (String, "Quit", "Save + Quit") var quit_text: String setget set_quit_text
 
+var _old_focus_owner: Control
```
Poobslag, line 13: Worth a comment.

src/main/ui/menu/system-buttons.gd
```diff
@@ -3,10 +3,33 @@ extends VBoxContainer
 Navigational buttons which occur on most menu screens.
 """
 
+export (bool) var _quit_on_cancel: bool = true setget set_quit_on_cancel
```
Poobslag, line 6: Comment this property.

Poobslag: Added comment, `If true, the 'ui_cancel' action...`

src/main/ui/menu/settings-menu.gd
Poobslag: Added comment, `the UI control which...`

PR #323, opened 2020-06-07
High scoring losses no longer wipe out low scoring victories

Poobslag:
This involved a save change, so the save version has been bumped up from
15d2 to 163e


src/main/old-save.gd
```diff
@@ -80,6 +80,90 @@ func transform_old_save() -> void:
...
+"""
+func attach_time_under_success(save_item: SaveItem, target_time: float) -> void:
+	for rank_result_obj in save_item.value:
+		var rank_result_dict: Dictionary = rank_result_obj
+		rank_result_dict["success"] = not rank_result_dict.get("lost", true) \
+				and rank_result_dict.get("seconds", 0.0) <= target_time
+
+
 """
 Appends a 'compare' flag to the ultra records.
```
Poobslag, line 168: It might be good to flag these changes with the version they're modifying (15d2, 0517)

Poobslag: Method names are now suffixed with `for_15d2`. The other 0517 method was already suffixed.

PR #346, opened 2020-06-12
Added 2D environment, floor, walls, paths, grass.

Poobslag:


assets/main/world/environment/path-00-sheet.png.import
```diff
@@ -0,0 +1,34 @@
+[remap]
```
Poobslag, line 1: This asset could have a better name. Maybe 'path-stone-sheet' after the path type.

src/main/world/environment/grass-tufts.gd
```diff
@@ -0,0 +1,46 @@
+extends YSort
+"""
+Generates grass tufts over grassy parts of a tilemap.
+"""
+
+const GRASS := 22
```
Poobslag, line 6: `GRASS_CELL` would be more descriptive.

assets/main/world/environment/path-00-sheet.png.import
Poobslag: Renamed to `path-stone-sheet.png`

src/main/world/environment/grass-tufts.gd
Poobslag: Renamed to `GRASS_CELL`

Issue #361, opened 2020-06-14
Restore spatial creature voices


Poobslag:
Creature voices like 'hello', and 'goodbye' used to have a spatial element. This was rolled back as a part of making the RestaurantScene use Creature2Ds on June 13th, 2020. We should try to bring back the spatial voices.

The most logical way would be for the voice to be a part of the Creature2D, instead of a part of the Creature.

Poobslag:
When restoring panning, we should also address or reopen #213.

PR #362, opened 2020-06-14
Simplified outline logic. RestaurantScene uses Customer2D.

Poobslag:
Instead of having a dozen outline sprites and shaders on Creature,
outlines are now applied to the viewport in Creature2D. This requires
fewer nodes, fewer shaders, and overall should perform better, require
less code, be easier to maintain, and be more reusable.

Customer food effects (head recoil, food noise) are now triggered by
AnimationPlayer

Changed sounds from AudioStreamPlayer2D to AudiostreamPlayer. Since these
are coming through a viewport now, their position is incorrect. To make
spatial sounds work correctly we most likely want to move these outside of
the viewport. I've opened issue #361 to capture this.

Added missing food laser for tentacle mouth.

Fixed conflicts between movement and fat creatures. The movement logic was
constantly pulling the head back to 0,0 -- but fat creatures need their head to be elsewhere.


src/demo/ui/chat/MoodDemo.tscn
```diff
@@ -7,14 +7,15 @@
 script = ExtResource( 3 )
 
 [node name="ColorRect" type="ColorRect" parent="."]
+visible = false
```
Poobslag, line 10: Restore this background.

src/main/world/creature/creature.gd
```diff
@@ -77,6 +77,8 @@ var _suppress_sfx_signal_timer := 0.0
 # forces listeners to update their animation frame
 var _force_orientation_change := false
 
+var line_rgb: Color
```
Poobslag, line 80: Document this property

src/demo/ui/chat/MoodDemo.tscn
Poobslag: Restored

src/main/world/creature/creature.gd
Poobslag: Added comment, `the color of this creature's lines...`

PR #377, opened 2020-06-15
ResourceCache is multithreaded for faster performance. Crushed PNGs.

Poobslag:
The game starts up 220% faster on my desktop, after changing the startup
logic to use four different threads. I experimented with different thread
counts. With one thread startup takes 2200 ms. With four threads it takes 1000
ms. More than four threads doesn't improve performance.

Crushed chat PNGs. I wanted to see if it would help performance, but it
didn't.


src/main/resource-cache.gd
```diff
@@ -19,17 +22,19 @@ export (bool) var minimal_resources := false
 
 # maintains references to all resources to prevent them from being cleaned up
 var _cache := {}
+var _cache_mutex := Mutex.new()
 
 # setting this to 'true' causes the background thread to terminate gracefully
 var _exiting := false
 
 # background thread for loading resources
```
Poobslag, line 30: `background threads`

src/main/resource-cache.gd
```diff
@@ -85,9 +92,13 @@ func _preload_all_pngs(userdata: Object) -> void:
...
 func _preload_next_png() -> void:
-	_load_resource(_remaining_resource_paths.pop_front())
+	_remaining_resource_paths_mutex.lock()
+	var path: String = _remaining_resource_paths.pop_front()
+	_remaining_resource_paths_mutex.unlock()
+	
+	_load_resource(path)
 	_work_done += 1.0
-	if not _remaining_resource_paths:
+	if is_done():
```
Poobslag, line 101: This signal could theoretically be emitted by multiple threads.

src/main/resource-cache.gd
Poobslag: Changed to 'background threads'

src/main/resource-cache.gd
Poobslag: Added mutex when emitting final signal

PR #381, opened 2020-06-16
Fixed bug where failing a rank mode counted as clearing it.

Poobslag:
Levels have independent 'lose' and 'success' conditions; you might win if
you get 4,000 points, but lose if you top out. However, it's possible to
do both.

I've changed it so the loss condition overrides the success condition;
success is never 'true' if you lose.

Reformatted test saves as json to be more human readable.


src/main/old-save.gd
```diff
@@ -115,32 +117,55 @@ Transforms the specified json save items to the latest format.
...
-						"rank-8d": _append_score_success_for_15d2(save_item, 4000)
-						"rank-9d": _append_score_success_for_15d2(save_item, 6000)
-						"rank-10d": _append_score_success_for_15d2(save_item, 15000)
-				"version":
-					json_save_item_obj["value"] = PlayerSave.PLAYER_DATA_VERSION
+	match version_string:
+		"15d2":
+			json_save_items = _convert_15d2(json_save_items)
+		"163e":
+			json_save_items = _convert_163e(json_save_items)
```
Poobslag, line 124: Be consistent; newest to oldest.

Poobslag: Reordered; 163e now precedes 15d2.

PR #389, opened 2020-06-19
Moved Puzzle hud items into CanvasLayer. Tutorial restart fix.

Poobslag:
Puzzle now collects its hud items (chalkboard, messages, buttons) in a
PuzzleHud CanvasLayer. This ensures they're not overwritten by visual
effects like lights, and also makes it easier to add future HUD items (e.g
mobile controls)

If the user got to the last "Earn $100" part of the tutorial and hit
'ESC', they would be taken to the ultra-normal scenario. I've changed it
so that they're taken to the start of the tutorial.


src/main/puzzle/tutorial/skill-tally-item.gd
```diff
@@ -62,14 +63,10 @@ func increment() -> void:
 	_update_label()
 
 
-"""
-Connects the signals in 'signal_names' to the appropriate puzzle nodes.
-"""
-func _connect_signals() -> void:
-	if Engine.is_editor_hint():
-		return
```
Poobslag, line 70: Preserve this comment; `Connects the signals in 'signal_names' to the appropriate puzzle nodes.`

src/main/puzzle/tutorial/skill-tally-item.gd
```diff
@@ -8,16 +9,14 @@ This is used during tutorials so the player can see what to do.
 # if 'true' the skill tally will be shown as '58%' instead of '52/90'
 export (bool) var show_as_percent: bool
 
-# path of the Puzzle node to monitor
```
Poobslag, line 11: Preserve this comment, `the Puzzle to monitor`

Poobslag: Preserved comment, `puzzle to monitor for things such as moving the piece and clearing lines`

src/main/puzzle/tutorial/skill-tally-item.gd
Poobslag: Preserved comment; `Connects the signals in 'signal_names' to the appropriate nodes.`

PR #399, opened 2020-06-20
Puzzle mode touchscreen controls now handle diagonal presses.

Poobslag:
Removed old TouchControlSet, TouchControlButton. The idea of using four
TouchscreenButtons for a controller-like input was the easiest thing, but
it does not work very well. The EightWay component works much more
effectively.


src/main/ui/touch/eight-way.gd
```diff
@@ -6,8 +6,8 @@ The player can press each button individually to emit different actions. If diag
 can press two diagonally adjacent buttons with a single touch.
 """
 
-# radius from the eightway's center where touches should be processed
-const RADIUS = 120
+# radius from the eightway's center where touches should be processed. significanly larger than its visual radius
```
Poobslag, line 9: typo; `significanly`

Poobslag:
Commit comment should clarify, 'for mobile'

src/main/ui/touch/eight-way.gd
Poobslag: Fixed; `significantly`

PR #402, opened 2020-06-21
Control improvements; left+right, ccw+cw. Hid puzzle buttons.

Poobslag:
Fixed left/right puzzle input issue. If you pressed left, then pressed right,
then released right -- the piece would not move left. While this isn't a
problem the way people usually play, it resulted in an unintuitive state where
you'd be holding the left arrow key with the piece stuck to the right wall.

Loco touchscreen controls no longer accept ui_left/ui_right together.

Pressing both rotate buttons simultaneously flips a piece.

Hid puzzle screen touch buttons. These were always visible, but they
should only be visible in touch mode.

### DIFF
--- a/src/main/puzzle/piece/piece-type.gd
+++ b/src/main/puzzle/piece/piece-type.gd
@@ -26,11 +26,14 @@ var cw_kicks: Array
 # array of piece kicks to try when rotating counterclockwise
 var ccw_kicks: Array
 
+# position the piece should kick to when flipping in place
+var flips: Array
+
 # maximum number of 'floor kicks', kicks which move the piece upward
 var max_floor_kicks: int
 
-func _init(init_string: String, init_pos_arr: Array, init_color_arr: Array, init_kicks: Array,
-		init_max_floor_kicks := 3) -> void:
+func _init(init_string: String, init_pos_arr: Array, init_color_arr: Array,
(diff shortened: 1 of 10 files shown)


src/main/puzzle/piece/piece-type.gd
```diff
@@ -26,11 +26,13 @@ var cw_kicks: Array
 # array of piece kicks to try when rotating counterclockwise
 var ccw_kicks: Array
 
+var flips: Array
```
Poobslag, line 29: Document this field.

src/main/puzzle/piece/piece-types.gd
```diff
@@ -121,7 +123,8 @@ var piece_o := PieceType.new("o",
 		# color data
 		[[Vector2(10, 3), Vector2(6, 3), Vector2(9, 3), Vector2(5, 3)],
 		[Vector2(10, 3), Vector2(6, 3), Vector2(9, 3), Vector2(5, 3)]],
-		KICKS_NONE
+		KICKS_NONE,
+		[Vector2(0, 0), Vector2(0, 0), Vector2(0, 0), Vector2(0, 0)]
```
Poobslag, line 127: These values seem like they could be calculated. Maybe something for refactoring day.

src/main/puzzle/piece/piece-type.gd
Poobslag: Added comment.

PR #425, opened 2020-06-27
Removed frosting pool. Removed 'remove_child' calls.

Poobslag:
Pooling frosting globs required us to move them back and forth between
several parent nodes. The remove_child calls required when moving a glob
from one parent to another sporadically altered the 2D render order because of
Godot #16213 (https://github.com/godotengine/godot/issues/16213).
This caused an unpleasant flickering effect in the frosting.

Removing the object pool addressed the issue. I've also removed all other
remove_child calls throughout the project. queue_free() will remove
children.

### DIFF
--- a/src/main/puzzle/frosting-globs.gd
+++ b/src/main/puzzle/frosting-globs.gd
@@ -7,13 +7,13 @@ direction.
 """
 
 # emitted when a frosting glob hits a wall
-signal hit_wall(glob, glob_alpha)
+signal hit_wall(glob)
 
 # emitted when a frosting glob smears against the playfield, next pieces, or gutter.
 # passes in a copy of the original frosting glob, because the original remains unchanged
-signal hit_playfield(glob_copy, glob_alpha)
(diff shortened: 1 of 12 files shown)


src/main/puzzle/frosting-globs.gd
```diff
@@ -7,52 +7,24 @@ direction.
...
 # emitted when a frosting glob smears against the playfield, next pieces, or gutter.
 # passes in a copy of the original frosting glob, because the original remains unchanged
-signal hit_playfield(glob_copy, glob_alpha)
-signal hit_next_pieces(glob_copy, glob_alpha)
-signal hit_gutter(glob_copy, glob_alpha)
+signal hit_playfield(glob)
+signal hit_next_pieces(glob)
+signal hit_gutter(glob)
+
+export (NodePath) var _playfield_path: NodePath
```
Poobslag, line 18: move exports after constants

Poobslag: moved

PR #432, opened 2020-06-29
Made setters consistent. Export variables are 'public'.

Poobslag:
All setters now follow the convention of 'set_foo(new_foo)'. This avoids
shadowing and complications related to setting public fields versus
private fields.

All export variables are now 'public', not prefixed with an underscore.
Any variable which is public enough to be set in the godot editor is
public enough for other classes to see them.

Moved delint.sh into project root.


project/src/main/ui/chat/ChatChoices.tscn
```diff
@@ -19,24 +19,21 @@ script = ExtResource( 2 )
...
 margin_left = 142.0
 margin_right = 280.0
 margin_bottom = 115.0
-_choice_text = "That's an interesting choice"
 
 [node name="Button2" parent="." instance=ExtResource( 6 )]
 margin_top = 119.0
 margin_right = 138.0
 margin_bottom = 234.0
-_choice_text = "I wouldn't choose this one!"
```
Poobslag, line 39: These shouldn't have been removed.

Poobslag: Restored `choice_text`.

PR #435, opened 2020-06-30
Improved Creature.tscn behavior in editor. Changed eye frames.

Poobslag:
Creature tool scripts can now be toggled with an `edit-creature.sh`
shell script. The creature's appearance can be manipulated and reset with
three new properties, 'reset_frames', 'reset_creature' and
'random_creature'.

Eyes now have a north-facing frame 2, just like mouths and other body
parts. This lets eyes reuse the same 'CreaturePackedSprite' logic as all
other body parts. Eyes aren't visible when facing away, but eventually they
might be. (When that happens we'll need them on a different Z-layer than they
are now.)

Split 'fill_creature_def' function out of 'summon'. 'Summon' is now just
'set_creature_def'.

Minor refactoring. Fixed inconsistencies in shell scripts (unquoted
strings, header style). Renamed inaccurate 'creature' fields to
'creature_visuals'.

### DIFF
--- a/delint.sh
+++ b/delint.sh
@@ -1,5 +1,6 @@
 #!/bin/sh
-# analyzes code for stylistic errors, printing them to the console
+################################################################################
+# Analyzes code for stylistic errors, printing them to the console
 
 # functions missing return type
 grep -R -n "^func.*):$" --include="*.gd" ./project/src
@@ -17,8 +18,11 @@ grep -R -n "var [^:]* = " --include="*.gd" ./project/src \
   | grep -v " = parse_json(" \
   | grep -v "chat-event.gd:73"
 
-find ./project/src -name *.TMP
-find ./project/src -name *.gd~
+find ./project/src -name "*.TMP"
+find ./project/src -name "*.gd~"
 
 # project settings which are enabled temporarily, but shouldn't be pushed
 grep "emulate_touch_from_mouse=true" ./project/project.godot 
+
+# check for enabled creature tool scripts; these should be disabled before merging
+grep -lR "^tool #uncomment to view creature in editor" project/src/main/world/creature
--- a/edit-creature.sh
+++ b/edit-creature.sh
@@ -1,0 +1,27 @@
+#!/bin/sh
+################################################################################
+# This script toggles the 'tool' keyword for our creature scripts as a
+# workaround for Godot #40000
+# (https://github.com/godotengine/godot/issues/40000)
+#
+# edit-creature.sh on: Enable creature editing in the editor. This causes errors
+#     in the Godot console.
+#
+# edit-creature.sh off: Disable creature editing in the editor. This fixes the
+#     errors in the Godot console.
+
+tool_line='tool #uncomment to view creature in editor'
+
+if [ "$1" = "on" ]
+then
+  # uncomment 'tool' lines to allow editing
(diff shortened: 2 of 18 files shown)


delint.sh
```diff
@@ -17,8 +18,10 @@ grep -R -n "var [^:]* = " --include="*.gd" ./project/src \
...
 
-find ./project/src -name *.TMP
-find ./project/src -name *.gd~
+find ./project/src -name "*.TMP"
+find ./project/src -name "*.gd~"
 
 # project settings which are enabled temporarily, but shouldn't be pushed
 grep "emulate_touch_from_mouse=true" ./project/project.godot 
+
+grep -lR "^tool #uncomment to view creature in editor" project/src/main/world/creature
```
Poobslag, line 27: This could use a comment.

edit-creature.sh
```diff
@@ -0,0 +1,27 @@
...
+#     in the Godot console.
+#
+# edit-creature.sh off: Disable creature editing in the editor. This fixes the
+#     errors in the Godot console.
+
+tool_line='tool #uncomment to view creature in editor'
+
+if [ "$1" = "on" ]
+then
+  # replace 'tool #uncomment' with '#tool #uncomment'
```
Poobslag, line 17: This is hard to read for humans. Maybe something more like `# comment out tool lines`

project/src/main/packed-sprite.gd
```diff
@@ -79,6 +79,8 @@ Loads the Aseprite json file.
 Stores the offset and region_rect for each frame.
 """
 func _load_frame_data() -> void:
+	_frame_src_rects.clear()
+	_frame_dest_rects.clear()
```
Poobslag, line 83: This is worth mentioning in the commit comment.

project/src/main/world/creature/mouth-anims.gd
```diff
@@ -8,42 +7,28 @@ An AnimationPlayer which animates mouths.
...
 
-"""
-Updates the frame to something appropriate for the creature's orientation.
-
-Usually the frame is controlled by an animation. But when it's not, this ensures it still looks reasonable.
-"""
-func _apply_default_frames() -> void:
-	if Engine.is_editor_hint():
-		_apply_tool_script_workaround()
-	_mouth.frame = 1 if _creature.orientation in [Creature.SOUTHWEST, Creature.SOUTHEAST] else 2
```
Poobslag, line 44: After removing the last reference to `_mouth`, you should remove the field too

delint.sh
Poobslag: Added comment, ` # check for enabled creature tool scripts; these should be disabled before merging`

project/src/main/world/creature/mouth-anims.gd
Poobslag: Removed unused `_mouth` field.

project/src/main/packed-sprite.gd
Poobslag: Expanded commit comment, `Fixed bug where setting PackedSprite.frame_data to null did not empty the
cached sprite areas.`

edit-creature.sh
Poobslag: Rephrased to `# uncomment 'tool' lines to allow editing`

PR #437, opened 2020-07-01
Added close button for mobile. Fixed export preset script.

Poobslag:
Export preset script was inadvertently broken when extracting the
'project' folder.


project/src/main/puzzle/puzzle-touch-buttons.gd
```diff
@@ -46,6 +46,11 @@ const SCHEMES := {
 # if false, pressing the buttons won't emit any actions.
 export (bool) var emit_actions := true setget set_emit_actions
 
+onready var _close := preload("res://assets/main/ui/touch/close.png")
+onready var _close_pressed := preload("res://assets/main/ui/touch/close-pressed.png")
+onready var _duck := preload("res://assets/main/ui/touch/duck.png")
+onready var _duck_pressed := preload("res://assets/main/ui/touch/duck-pressed.png")
```
Poobslag, line 52: Maybe worth commenting why there's a duck in here.

Poobslag: Added comment, `when the player is testing buttons, we replace the icon so we don't confuse users trying who are trying to quit`

PR #438, opened 2020-07-02
Moved creature data from scenes into json files.

Poobslag:
Creatures now have a creature.json which points to dialog in the same
path. This modular design makes it easy to add more creatures.

Widespread renames:
  * creature_def -> dna
  * chat_id -> id, creature_id
  * chat_name -> name, creature_name, chat_name

Appearance is only one small part of a 'creature definition' -- a creature
definition also includes things like their name, their chat window
appearance, and where their dialog is stored. The old 'creature_def' which
specifically determined their appearance has been renamed to 'dna'.


project/src/main/ui/chat/chat-library.gd
```diff
@@ -10,39 +10,25 @@ into the UI.
...
-	var chat_tree
-	var focused: Node = ChattableManager.get_focused()
-	if not focused.has_meta("chat_id"):
-		# can't look up chat events without a chat_id; return an empty array
-		push_warning("Chattable %s does not define a 'chat_id' property." % focused)
-	else:
-		# open the json file for the currently focused chatter
-		var chat_id: String = focused.get_meta("chat_id")
-		chat_tree = load_chat_events_from_file("res://assets/main/dialog/%s.json" % chat_id)
-	return chat_tree
```
Poobslag, line 28: This moved method is worth describing in a commit comment.

project/src/main/ui/chat/chat-library.gd
```diff
@@ -10,39 +10,25 @@ into the UI.
...
+static func creature_def(id: String) -> CreatureDef:
+	var creature_def_text: String = FileUtils.get_file_as_text("res://assets/main/dialog/%s/creature.json" % id)
+	var json_creature_def: Dictionary = parse_json(creature_def_text)
+	var creature_def := CreatureDef.new()
+	creature_def.from_json_dict(json_creature_def)
+	return creature_def
```
Poobslag, line 18: `load_creature_def` is a better name

project/src/main/ui/chat/chat-library.gd
Poobslag: Commit comment now mentions `Moved load_chat_events function from ChatLibrary into ChattableManager`

project/src/main/ui/chat/chat-library.gd
Poobslag: Renamed and moved into `CreatureLoader`

PR #452, opened 2020-07-06
Split out CreatureVisuals again. Fixed disappearing arms.

Poobslag:
These were combined to avoid cumbersome bugs because FoodLaser and sound
effects were behind a viewport. However, having this in a single scene
prevents Godot's path editor from working properly which makes the
upcoming creature changes (shadows, patterns) cumbersome to work with.

The creature's arms were disappearing because of signal disconnected by 6cc35649.


project/src/main/world/creature/creature-outline.gd
```diff
@@ -0,0 +1,16 @@
+extends Node2D
```
Poobslag, line 1: Delete this.

project/src/main/world/creature/smooth-path-outline.gd
```diff
@@ -1,13 +0,0 @@
-tool
-extends SmoothPath
```
Poobslag, line 2: Document this in the commit comment.

project/src/main/world/creature/creature-outline.gd
Poobslag: Deleted.

project/src/main/world/creature/smooth-path-outline.gd
Poobslag: Added comment, `Removed unused smooth-path-outline.gd script.`.

Issue #454, opened 2020-07-10
Shadows should work when creatures face north


Poobslag:
Currently shadows look OK when fatter creatures face south, but not when they turn to face north.

This never happens in the game currently, but it might soon.

Poobslag:
Shadows look good enough for now. There's technically a missing shadow under their arm, but it's passable.

PR #455, opened 2020-07-10
Fat creatures have better shadows.

Poobslag:
The body shadows are implemented as a ViewportContainer (!?) for a very
bizarre reason having to do with Godot's limitations regarding drawn
nodes, shaders, textures, and a few other things. But to draw these
shadows over our creature body, we need to apply a shader to a custom
drawn node, which does not on the surface sound very difficult.

The first idea one might have for applying a shader to a custom drawn
node, is to apply shaders to a Node2D which has some custom draw_ commands.
However due to Godot #29181 (among other things), shaders cannot be applied to
custom drawn nodes. These nodes do not have UVs assigned, so textures will
report all of their pixels as white.

The second idea one might have is complicate things by using a Node2D,
shader, and a Texture. The Node2D can be drawn to a texture, and the
texture can have a shader on it. However Godot does not provide any way to
draw directly to a texture without leveraging viewports.

A third idea one might have is to complicate things further by using a
Node2D, Shader, and a ViewportTexture which leverages a Viewport. The
Node2D can be drawn into a Viewport, the ViewportTexture leverages that
Viewport and the ViewportTexture has a shader on it. This works but it
interferes with the Godot editor. We have Path2D objects with points and
handles we want to drag in the Godot editor, but Viewports render the
nodes invisible so the Godot editor no longer lets you manipulate them.

A fourth idea one might have is to make things absurdly complicated by using a
Node2D, Shader, and ViewportContainer which contains the Viewport. The Node2D
can be drawn into a Viewport. The Viewport is within a ViewportContainer
allowing the Godot editor to render it, and the ViewportContainer has a shader.
That mostly works -- however, Godot will also render the ViewportTexture at
runtime. If ViewportTexture.visible is set to false to prevent it from
rendering, then the Viewport will not be drawn. (render_target_update_mode does
not affect this).

A fifth idea one might have is to make things really absurdly complicated by
using a Node2D, Shader, a pseudo-visible ViewportContainer which contains the
Viewport. The Node2D can be drawn into a Viewport. The Viewport is within a
ViewportContainer allowing the Godot editor to render it, and the
ViewportContainer also has a shader. ViewportContainer.visible is set to true so
that the Godot editor THINKS it is being rendered and will provide the necessary
UI components, but its modulate property is set with an alpha component of 0 so
the player can't actually see it. This convoluted solution not only seems to
work, but it seems to be the simplest approach (!?)

--------------------------------------------------------------------------------

Added tool for creating shadow positions in editor

Minor improvements to fat creature appearance when facing north; still looks
very glitchy, I've opened #454 as it is outside the scope of this pull
request

Fixed bugs where SmoothPath didn't redraw for color changes


project/src/main/world/creature/fat-player.gd
```diff
@@ -6,6 +6,8 @@ plays an animation forward and backwards to adjust the creature's appearance.
 
 # how large the creature has grown; 5.0 = 5x normal size
 var _fatness := 1.0 setget set_fatness, get_fatness
+var _creature_orientation := CreatureVisuals.SOUTHEAST
```
Poobslag, line 9: add a comment for this -- not necessarily what it does, but why it's necessary

project/src/main/world/creature/smooth-path.gd
```diff
@@ -13,12 +13,17 @@ export(bool) var _smooth: bool setget smooth
...
 export(bool) var closed := true
 
-export(Color) var line_color := Color.black
+export(Color) var line_color := Color.black setget set_line_color
 export(Color) var fill_color := Color.transparent setget set_fill_color
 export(float) var line_width := 8.0
 
 # internal array used for drawing polygons
 var _poly_colors := PoolColorArray()
+var _poly_uvs := PoolVector2Array()
```
Poobslag, line 22: undo this 'poly_uvs' stuff

project/src/main/world/creature/fat-player.gd
Poobslag: Added comments, `we track where...`

project/src/main/world/creature/smooth-path.gd
Poobslag: Removed

Issue #461, opened 2020-07-13
Creature's 'angry cloud' emote doesn't look good when the creature is fat


Poobslag:
The 'angry cloud' emote should have different versions for fatter creatures.

Currently, it shows up as tiny clouds in the middle of their stomach which is strange. It should be bigger and lower.

Poobslag:
Alternatively, creatures who exceed a certain threshold could just have a totally different version of the emote, or no clouds at all.

PR #462, opened 2020-07-13
Fat creatures can now appear on the overworld.

Poobslag:
Eventually, creatures will show up fatter as you feed them more. For now
it's just a hard-coded value in the creature definition.

Most emotes work for fat creatures. Their arm shadow disappears
appropriately when they bring their arms up to their face. The 'rage cloud'
emote doesn't look very good, but that's a hard one to fix. I opened #461 to
document this.

### DIFF
--- a/project/src/main/world/chat-icon.gd
+++ b/project/src/main/world/chat-icon.gd
@@ -27,6 +27,9 @@ export (BubbleType) var bubble_type: int setget set_bubble_type
 # 'true' if this bubble should appear on the right side, 'false' if it's on the left side
 export (bool) var right_side := true setget set_right_side
 
+# the position the bubble should point to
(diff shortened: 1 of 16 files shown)


project/src/main/world/chat-icon.gd
```diff
@@ -27,6 +27,8 @@ export (BubbleType) var bubble_type: int setget set_bubble_type
 # 'true' if this bubble should appear on the right side, 'false' if it's on the left side
 export (bool) var right_side := true setget set_right_side
 
+export (Vector2) var target_position: Vector2 setget set_target_position
```
Poobslag, line 30: Document this new property

Poobslag: Added comment

PR #477, opened 2020-07-21
Ebe runs back and forth on overworld. CreatureVisuals improvements.

Poobslag:
If a creature did the 'think0' emote while facing north, their eyes
wouldn't be reset, so they'd have invisible eyes

Spira was emitting an 'update_orientation' signal every frame when running
north.


project/project.godot
```diff
@@ -387,7 +387,7 @@ _global_script_classes=[ {
 "base": "Creature",
 "class": "Spira",
 "language": "GDScript",
-"path": "res://src/main/world/spira.gd"
+"path": "res://src/main/world/creature/spira.gd"
```
Poobslag, line 390: Mention this move in the commit comment.

todo.txt
```diff
@@ -0,0 +1,22 @@
+%modify FSM to remove 'frame' stuff
```
Poobslag, line 1: remove

project/project.godot
Poobslag: Added comment, `Moved spira.gd from /world into /world/creature`

todo.txt
Poobslag: removed

PR #481, opened 2020-07-24
Rolling backup saves.

### LINKED ISSUE #122: Add rolling backup saves
Currently, we load save data from something like "file.dat" and save it back to "file.dat". This work OK but a failure will wipe out all of the user's save data. As someone who suffered through replaying the first 30 minutes of Outer Wilds on the PS4 twenty times because of save corruption, I'd like to make our save system more resilient.  We should consider the following scenarios:

  * What if a bug in the game writes weird save data, like writing a string into a field that should be numeric?
  * What if a bug in the game makes it crash while writing save data, so that only half of a save file gets written?
  * What if the user's operating system temporarily or permanently denies us access to a file?

The save process could keep track of old save data so the user can go back 1 hour, 1 day, 1 week or 1 month. It should also always keep the save with the longest playtime.

```
turbofat0.longest-playtime-sav.bak
turbofat0.this-hour.sav.bak
turbofat0.prev-hour.sav.bak
turbofat0.this-day.sav.bak
turbofat0.prev-day.sav.bak
turbofat0.this-week.sav.bak
turbofat0.prev-week.sav.bak
```

When the player saves:
  * Check if the **this-hour** backup is at least an hour old
    * If so, copy it to the **prev-hour** backup, and copy the current save as the **this-hour** backup
  * Do the same for the other durations

When the player loads:
  * If a problem is detected, try loading the backups (newest to oldest) until one works
  * Notify the player that you've rolled back because there were problems.

Additionally, when save slots are introduced, we might have options for "new game" but also "restore an old save"

Poobslag:
The game keeps a backup for the past hour/day/week. If the newest save
file can't be loaded, it loads a backup and prints a message

Closes #122.

### DIFF
--- a/project/src/main/ui/menu/bad-save-data-control.gd
+++ b/project/src/main/ui/menu/bad-save-data-control.gd
@@ -1,0 +1,58 @@
+extends ColorRect
+"""
+A popup window which is shown if there is a problem loading the save data.
+"""
+
+func _ready() -> void:
+	hide()
+
+
+"""
+Displays a popup window describing the problem loading the save data.
+
+The popup window is populated based on PlayerSave's current state.
+"""
+func popup() -> void:
+	show()
+	$Popup.popup()
+	
+	# Construct a message to show the player. The tone is informal and sympathetic. Hopefully this endears the player
+	# toward us because they're going to have a bad day. (Hopefully the backups worked.)
+	var message := ""
+	if PlayerSave.loaded_rolling_save == -1:
+		message += "So... I had some kind of problem loading your save data."
+		message += " Unfortunately, none of the backups worked either, so I don't really know how to fix it."
+	else:
+		message += "So... I had some kind of problem loading your save data,"
+		message += " which means you're going to lose some progress."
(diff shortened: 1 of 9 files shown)


project/src/main/ui/menu/bad-save-data-control.gd
```diff
@@ -0,0 +1,57 @@
...
+	$Popup.popup()
+	
+	# Construct a message to show the player. The tone is informal and sympathetic. Hopefully this endears the player
+	# toward us because they're going to have a bad day. (Hopefully the backups worked.)
+	var message := ""
+	if PlayerSave.loaded_rolling_save == -1:
+		message += "So... I had some kind of problem loading your save data."
+		message += " Unfortunately, none of the backups worked either, so I don't really know how to fix it."
+	else:
+		message += "So... I had some kind of problem loading your save data, which means you're going to lose some progress."
```
Poobslag, line 26: line too long

Poobslag: Split line.

PR #482, opened 2020-07-24
Mobile controls appear over 'black bars'. Changed overworld scaling.

### LINKED ISSUE #390: Mobile controls should fill in the 'black borders' when resized
The mobile controls are currently in the center of the screen, but they should overlap the black area on the outside.

Otherwise, your thumbs need to stretch to the middle of your phone and it's less comfortable.

Poobslag:
This required rewriting a lot of the scaling code for things like
positioning buttons.

The overworld now scales to fill in the edges, so it doesn't have 'black
bars'. This required rewriting the shadow/background logic to expand
dynamically.

Closes #390.


project/src/main/ui/menu/scene-camera.gd
```diff
@@ -0,0 +1,17 @@
+extends Camera2D
```
Poobslag, line 1: Remove.

Poobslag: Removed.

PR #484, opened 2020-07-24
Audio feedback when testing out touch controls.

Poobslag:
...To be honest I just wanted an excuse for the duck button to make a
quack sound.


project/project.godot
```diff
@@ -682,6 +682,10 @@ soft_drop={
  ]
 }
 
+[input_devices]
+
+pointing/emulate_touch_from_mouse=true
+
```
Poobslag, line 688: Disable.

Poobslag: Disabled

Issue #467, opened 2020-07-18
Volume settings improvements


Poobslag:
There should be a 'master' volume setting as well.

These should all be categorized under 'Volume' so people know what they mean.

PR #487, opened 2020-07-24
New settings: Ghost piece, master volume. Back button fixes.

### LINKED ISSUE #485: Add support for an optional 'ghost piece'
This is at least helpful for beginners. It should not be mandatory.

Poobslag:
Ghost piece can be toggled with a setting. It defaults to 'on', and if
it's on a shadow appears at the piece's hard drop destination.

Master volume can be toggled, which affects the volume of
sfx/music/voices. I've changed the default volume from 0.5 to 0.7 across
the board, to compensate for the fact that all of the different audio
buses are affected by the master volume.

Fixed broken back button shortcuts. These were broken during the recent
touchscreen UI changes.

Closes #485, #467.


project/src/main/puzzle/piece/piece-dropper.gd
```diff
@@ -11,6 +11,7 @@ export (NodePath) var piece_mover_path: NodePath
 
 # 'true' if the player hard dropped the piece this frame
 var did_hard_drop: bool
+var hard_drop_target_pos: Vector2
```
Poobslag, line 14: Add a comment

project/src/main/puzzle/Playfield.tscn
```diff
@@ -182,7 +182,7 @@ script = ExtResource( 5 )
 puzzle_tile_map_path = NodePath("../TileMap")
 
 [node name="ShadowTexture" type="TextureRect" parent="."]
-modulate = Color( 1, 1, 1, 0.196078 )
+modulate = Color( 1, 1, 1, 0.313726 )
```
Poobslag, line 185: Document why this value was changed.

project/src/main/puzzle/puzzle-tile-map.gd
```diff
@@ -26,6 +26,9 @@ const COL_COUNT = 9
 # a number in the range [0, 1] which can be set to make the tile map flash or blink.
 var whiteness := 0.0 setget set_whiteness
 
+# Pieces have a second shadow at the destination
```
Poobslag, line 29: Mention "ghost" somewhere in the comment, and maybe change the field name as well to something like "ghost shadow offset"

project/src/main/ui/settings-ghost-piece.gd
```diff
@@ -0,0 +1,8 @@
+extends Control
```
Poobslag, line 1: Add comment.

project/src/main/puzzle/piece/piece-dropper.gd
Poobslag: Added comment, `The hard drop destination for the current piece...`

project/src/main/puzzle/Playfield.tscn
Poobslag: Added commit message, `Made piece shadows slightly darker so that ghost piece would be more
visible.`

project/src/main/puzzle/puzzle-tile-map.gd
Poobslag: Changed documentation to `offset used for drawing the 'ghost piece'`, and changed variable name to include `ghost`

project/src/main/ui/settings-ghost-piece.gd
Poobslag: Added

PR #488, opened 2020-07-27
Added tutorial button. Fixed tutorial star seed bug.

Poobslag:
Added a tutorial button to the main menu for testing tutorial
functionality. This is a placeholder, and will eventually behave
differently and possibly not be located on the main menu.

The tutorial's star seeds were lingering from one part of the tutorial to
the next. I've removed the logic for spawning the star seeds at the start of a
level because it was the easiest way to prevent these bugs.


project/assets/main/music/hiphop-08.ogg.import
```diff
@@ -0,0 +1,15 @@
+[remap]
```
Poobslag, line 1: Remove these files.

Poobslag: Removed.

PR #506, opened 2020-08-04
Tutorial bugs: star seeds, instructor, music volume, messages

Poobslag:
Fixed bug where star seeds sometimes lingered during tutorial. This bug was
caused because the after_piece_written and lines_deleted signals were emitted in
the wrong order. This meant when changing tutorial scenes, the scene would
change, and then the new star seeds would get moved around based on the cleared
lines.

Star seeds for incomplete boxes (such as in the third part of the tutorial) are
no longer placed adjacent if possible.

Instructor is now summoned when restarting a tutorial. Before, if you restarted
a tutorial after customers showed up, the random customer would lead the tutorial.

Increased tutorial music volume. The tutorial music is about 4 db quieter
than other songs, so I've turned up its in-game volume to compensate.
It was kind of jarring when the 'real music' kicked in before, because it
was a lot louder.

Fixed bug where message would sometimes be blanked out when restarting a
tutorial. This was because the pop_out_timer wasn't reset, so a message
might be marked to pop out, and then an important message appears
but disappears before it can be read.

### DIFF
--- a/project/src/main/music-player.gd
+++ b/project/src/main/music-player.gd
@@ -110,8 +110,17 @@ Gradually fades a track in.
 Usually it's OK for a track to start abrubtly, but sometimes we want to fade music in more gradually.
 """
 func fade_in() -> void:
-	current_bgm.volume_db = -40.0
-	$MusicTween.fade_in(current_bgm)
+	current_bgm.volume_db = MIN_VOLUME
+	$MusicTween.fade_in(current_bgm, max_volume(current_bgm))
+
+
+"""
+Returns the volume a song should play at.
+
+The tutorial music is noticably quieter than the other songs, so we increase its volume.
+"""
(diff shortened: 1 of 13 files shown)


project/src/main/music-player.gd
```diff
@@ -110,8 +110,17 @@ Gradually fades a track in.
...
 """
 func fade_in() -> void:
-	current_bgm.volume_db = -40.0
-	$MusicTween.fade_in(current_bgm)
+	current_bgm.volume_db = MIN_VOLUME
+	$MusicTween.fade_in(current_bgm, max_volume(current_bgm))
+
+
+"""
+Returns the volume a song should play at.
```
Poobslag, line 118: You shouldn't end a sentence with a preposition, Aaron.

Poobslag: ...Somehow I won't lose sleep over this.

PR #520, opened 2020-08-09
Added main character to restaurant scene.

Poobslag:
CreatureView is now RestaurantView. CreatureView had a parent/child relationship
with its RestaurantScene and did not use a Viewport or Camera. It's been
rewritten to RestaurantView which utilizes viewports and cameras so that we can
have multiple viewports into the same scene.

Despite utilizing viewports, the quality is still the same as it was
before even at high resolutions. This was achieved by having
Viewports rendering things at 200% size, which are then scaled down in
ViewportContainers. We should be able to use the same technique to improve
the quality of shadows and puzzle fruits later.

Chef emotes when player does well, tops out, or finishes a level.


project/src/main/puzzle/puzzle-score.gd
```diff
@@ -26,9 +26,11 @@ signal after_game_ended
 
 signal level_index_changed(value)
 
+signal added_line_score(combo_score, box_score)
 signal score_changed
 
 signal combo_ended
+signal topped_out
```
Poobslag, line 33: Preserve `# emitted when the current piece can't be placed in the _playfield` comment

project/src/main/puzzle/restaurant-view.gd
```diff
@@ -0,0 +1,112 @@
+extends Control
+"""
+Showing the player's chef character and active customer in a restaurant scene.
+
+As the player drops blocks and scores points, the characters animate and react.
+"""
+
+var current_creature_index: int setget set_current_creature_index, get_current_creature_index
+var _recent_bonuses := []
```
Poobslag, line 9: Document this 'recent bonuses' field, it's a weird one

project/src/main/world/creature/emote-anims.gd
```diff
@@ -42,12 +42,24 @@ const EAT_SWEAT_ANIMS := {
...
+	[ChatEvent.Mood.CRY0, ChatEvent.Mood.CRY0]: "_transition_noop",
+	[ChatEvent.Mood.CRY1, ChatEvent.Mood.CRY1]: "_transition_noop",
 	[ChatEvent.Mood.SWEAT0, ChatEvent.Mood.SWEAT1]: "_transition_noop",
+	[ChatEvent.Mood.SWEAT0, ChatEvent.Mood.SWEAT0]: "_transition_noop",
 	[ChatEvent.Mood.SWEAT1, ChatEvent.Mood.SWEAT0]: "_transition_sweat1_sweat0",
-	[ChatEvent.Mood.SMILE0, ChatEvent.Mood.SMILE1]: "_transition_noop",
-	[ChatEvent.Mood.SMILE1, ChatEvent.Mood.SMILE0]: "_transition_smile1_smile0",
+	[ChatEvent.Mood.SWEAT1, ChatEvent.Mood.SWEAT1]: "_transition_noop",
+	[ChatEvent.Mood.RAGE0, ChatEvent.Mood.RAGE0]: "_transition_noop",
+	[ChatEvent.Mood.RAGE1, ChatEvent.Mood.RAGE1]: "_transition_noop",
```
Poobslag, line 62: Document this change in the commit comment.

project/src/main/puzzle/puzzle-score.gd
Poobslag: Kept comment

project/src/main/puzzle/restaurant-view.gd
Poobslag: Added comment; ` # bonus points scored for recent lines`

project/src/main/world/creature/emote-anims.gd
Poobslag: Added comment: `Moods transition into themselves better`

PR #530, opened 2020-08-18
Added handkerchiefs.

Poobslag:
Creatures have a new 'collar' property. Currently the only choices are
'handkerchief' and 'off'.

Creatures have a new 'cloth_rgb' property. Rather than having overly
granular properties for headband color, collar color, glasses color --
there will be three overarching colors for hair, metal, cloth and glass.
This will apply to any clothing.


project/src/demo/ui/chat/MoodDemo.tscn
```diff
@@ -51,9 +54,11 @@ dna = {
 "line_rgb": "41281e",
 "mouth": "1",
 "nose": "3",
-"property:Body/Viewport/Body/NeckBlend:frame_data": "res://assets/main/world/creature/1/neck-packed.json",
+"property:Body/Viewport/Body/NeckBlend:frame_data": "res://assets/main/world/creature/1/neck-blend-packed.json",
```
Poobslag, line 57: Document this renaming in the commit comment.

Poobslag: Added comment, `avoid ambiguity with collar resources, as the collar is also neck-related`

PR #532, opened 2020-08-20
Added colored faces. Minor changes to eyes, cheeks, colors.

Poobslag:
Redrew eye assets to have a consistently opaque oval; otherwise they
resulted in some strange color patterns when overlapping cheeks. This
wasn't noticable when the cheeks and eyes were the same colors, but now
sometimes they're different.

Tweaked cheek shapes to have a rounded edge, so that they blend with
colored faces better

Changed some palettes. Changed belly colors to match body color closer;
some of these highly contrasting belly colors looked really bad on the
face. Also muted a few loud hair/eye colors, such as one creature's
radioactive purple eyes.


project/src/main/world/creature/CreatureVisuals.tscn
```diff
@@ -252,14 +328,90 @@ shader_param/green = Color( 0, 0, 0, 1 )
...
 resource_local_to_scene = true
-shader = ExtResource( 1 )
+shader = SubResource( 123 )
 shader_param/red = Color( 0, 0, 0, 1 )
 shader_param/green = Color( 0.952941, 0.572549, 0.454902, 1 )
 shader_param/blue = null
 shader_param/black = Color( 0, 0, 0, 1 )
 
+[sub_resource type="Shader" id=117]
+code = "/*
```
Poobslag, line 374: Do we need this same shader code repeated ~30 times? I know we need the parameters to be different but does the shader need to be different too?

Poobslag: The shader doesn't need to be different, just the material. I've removed this shader code.

PR #542, opened 2020-08-23
More points for building horizontally.

### LINKED ISSUE #540: Horizontal blocks should offer more points
The playfield is organized to make vertical blocks optimal. We should change the scoring so that horizontal rainbow blocks are tempting, even if they're still inefficient.

Instead of +10 points per row, a horizontal 4x3 could be +20 points per row (60 total), and a horizontal 5x3 could be +30 points per row (90 total.)

Poobslag:
The playfield is organized to make vertical blocks optimal. This commit
increases the score for horizontal boxes from +10 per row to +20 and +30
per row.

A horizontal 4x3 box used to score a total of 30 points; 25% less than a
vertical box. Now it scores 60 points; 50% more than a vertical box.

A horizontal 5x3 box used to score a total of 30 points; 40% less than a
vertical box. Now it scores 90 points; 80% more than a vertical box.

This also changes how star seeds are generated. Horizontal boxes include
more star seeds to help communicate the new scoring bonuses.

Closes #540.


project/src/main/puzzle/star-seeds.gd
```diff
@@ -93,37 +90,65 @@ func _add_wobblers_for_box(rect: Rect2, color_int: int) -> void:
 		wobbler_positions = []
 		var prev_position := randi() % int(rect.size.x)
```
Poobslag, line 91: `prev_position` is not referenced and can be removed.

Poobslag: Removed.

PR #547, opened 2020-08-27
anims -> player. Broke out creature animation players.

Poobslag:
Creature animation players are now in their own scenes.

Renamed EarAnims, MouthAnims, etc into EarPlayer, MouthPlayer for
consistency with AnimationPlayer. Since these are moving into their own
scenes I wanted them to have more consistent names.


project/project.godot
```diff
@@ -129,11 +129,6 @@ _global_script_classes=[ {
 "language": "GDScript",
 "path": "res://src/main/creature-library.gd"
 }, {
-"base": "Reference",
-"class": "CreatureLoader",
-"language": "GDScript",
-"path": "res://src/main/world/creature/creature-loader.gd"
-}, {
```
Poobslag, line 136: This probably doesn't belong in this commit.

Poobslag: Rolled back.

PR #557, opened 2020-08-31
Refactored CreatureVisuals methods, signals, repetitive code.

Poobslag:
Split CreatureVisuals._update_creature_properties into _unassign and
_assign methods. These are slightly more descriptive and help separate and
document the two distinct pieces of functionality the method was doing.

Extracted _add_visuals_node, _remove_visuals_node utility methods to cut
down on repetitive code.

The creature_arrived signal is now named 'dna_loaded' which is more
descriptive. This was originally named 'creature_arrived' because it coincided
with arriving at the restaurant, but now that is not always the case as it
applies to creatures being loaded in the overworld or elsewhere.

I removed the before_creature_arrived signal. It was only used for stopping
animations which is now unnecessary as the animation players get removed.

### DIFF
--- a/project/src/main/world/creature/creature-visuals.gd
+++ b/project/src/main/world/creature/creature-visuals.gd
@@ -18,11 +18,8 @@ unnecessary textures are loaded.
 # emitted on the frame when the food is launched into the creature's mouth
 signal food_eaten
 
-# emitted before a creature arrives and sits down
-signal before_creature_arrived
-
-# emitted when a creature arrives and sits down
-signal creature_arrived
+# emitted when a creature's textures and animations are loaded
+signal dna_loaded
 
 # emitted during the 'run' animation when the creature touches the ground
 # warning-ignore:unused_signal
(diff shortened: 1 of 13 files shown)


project/src/main/world/creature/creature-visuals.gd
```diff
@@ -353,28 +352,6 @@ func play_movement_animation(animation_prefix: String, movement_direction: Vecto
...
-	
-	# preserve the old orientation if it's close to the new orientation. this prevents us from flipping repeatedly
-	# when our direction puts us between two orientations.
-	var new_orientation: int = orientation
-	# unrounded orientation is a float in the range [-2.0, 2.0]
-	var unrounded_orientation := -2 * direction.angle_to(southeast_dir) / PI
-	if abs(unrounded_orientation - orientation) >= 0.6 and abs(unrounded_orientation + 4 - orientation) >= 0.6:
-		# convert the float orientation [-2.0, 2.0] to an int orientation [0, 3]
-		new_orientation = wrapi(int(round(unrounded_orientation)), 0, 4)
-	return new_orientation
```
Poobslag, line 375: Reordering these methods was good; reorder the smooth path methods too.

Poobslag: The smooth path methods appear to be in the correct order. It has some private/public intermingled fields, but there's an arguable purpose to that as it impacts the appearance of the editor.

PR #560, opened 2020-09-01
Upgrade Gut 6.8.1 -> 7.0.0

Poobslag:


project/src/test/tests.tscn
```diff
@@ -1,21 +1,43 @@
...
+_font_name = "AnonymousPro"
+_font_size = 20
+_font_color = Color( 0.8, 0.8, 0.8, 1 )
+_background_color = Color( 0.15, 0.15, 0.15, 1 )
+_color_output = true
+_select_script = ""
+_tests_like = ""
+_inner_class_name = ""
 _run_on_load = true
-_yield_between_tests = false
```
Poobslag, line 17: Tests run MUCH faster if this remains false. (0.0s instead of 1.1s)

Poobslag: Reset back to false.

PR #562, opened 2020-09-03
.tscn resources are cached, improving scene launch performance.

Poobslag:
Caching these resources improves the puzzle launch time from about 2
seconds to about 1 second.

Caching .tscn resources requires some special handling, since loading them
in a threaded manner causes problems. We need to load all of the resources
normally, and then load the scenes in the main thread.

Performance could be further improved by initializing and reusing a single
puzzle scene instead of initializing a new scene every time -- but this
introduces a lot of bugs reinitializing the scene, as well as errors/warnings
exiting the game.

### DIFF
--- a/project/src/main/resource-cache.gd
+++ b/project/src/main/resource-cache.gd
@@ -15,6 +15,9 @@ signal finished_loading
 # number of threads to launch; 1 is slower, but more than 4 doesn't seem to help
 const THREAD_COUNT := 4
 
+# loading scenes is slower than loading regular resources; this constant estimates how much slower
+const WORK_PER_SCENE := 8.0
+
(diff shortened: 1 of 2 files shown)


project/src/main/resource-cache.gd
```diff
@@ -14,6 +14,7 @@ signal finished_loading
 
 # number of threads to launch; 1 is slower, but more than 4 doesn't seem to help
 const THREAD_COUNT := 4
+const WORK_PER_SCENE := 3.0
```
Poobslag, line 17: This is worth commenting

Poobslag: Added comment; `this constant estimates how much slower`

PR #565, opened 2020-09-04
Reuse creature editor improvements elsewhere. BodyColors -> BellyColors.

Poobslag:
Reuse creature editor's dialog backdrop component in level editor. Level
editor's dialogs are now exclusive and grey out the level editor while
active.

Reuse creature editor's shadow logic in overworld, restaurant scene. This
allows scenes to automatically create shadows for any number of creatures,
rather than explicitly declaring Creature Shadow 1, Creature Shadow 2...

Renamed BodyColors to BellyColors for consistency with the creature
editor.


project/src/main/editor/puzzle/dialogs.gd
```diff
@@ -0,0 +1,28 @@
+extends Control
```
Poobslag, line 1: Add a class-level comment

Poobslag: Added comment; ` Shows popup dialogs for the level editor.`

PR #568, opened 2020-09-04
Replaced kebab case with snake case in JSON keys/values.

Poobslag:
This aligns with python standards. GDScript is not python, but it is similar in
structure.

Updated version numbers for dialog and scenarios. The snake case change
breaks backwards compatibility. I didn't implement any sort of backwards
compatibility logic because I strongly doubt anybody else has created their own
dialog/scenarios yet.


project/src/main/ui/overworld-ui.gd
```diff
@@ -102,7 +102,7 @@ func _process_scenario_meta_item(meta_item: String) -> void:
 
 
 """
-Process a 'select_level_*' event, loading the appropriate scenario data or conversation to launch.
+Process a 'select-level-*' event, loading the appropriate scenario data or conversation to launch.
```
Poobslag, line 105: should stay as underscores; not hyphens

Poobslag: Changed back to underscores.

PR #579, opened 2020-09-06
ChatIcons are generated for all chattables. Lowered squirrel chat icons.

Poobslag:
Creatures no longer have a chat icon -- instead, a separate ChatIcons node
generates chat icons for all creatures, and anything else that is in the
chattables group. Instead of customizing the chat icon by changing the
chat icon's properties, nodes can define metadata which affects their chat
icon appearance, such as the 'chat_bubble_type' metadata

This is a slightly cleaner design as scaling/rotating a creature won't
mess with their chat icon, and we don't need to worry about chat icons
being hidden by the environment.

These chat icons can't go in the UI layer because they're positioned using
RemoteTransform2D. But they're in a Node2D separate from the creatures and the
environment.

Lowered chat bubbles for small squirrel creatures. Their heads are lower,
so the chat icons should be lower as well.

### DIFF
--- a/project/src/main/world/glowy-sphere.gd
+++ b/project/src/main/world/glowy-sphere.gd
@@ -5,3 +5,4 @@ Script for showing a placeholder chattable object in the overworld.
 
 func _ready() -> void:
 	set_meta("chat_path", "res://assets/main/dialog/glowy-sphere.json")
+	set_meta("chat_bubble_type", ChatIcon.THOUGHT)
--- a/project/project.godot
+++ b/project/project.godot
@@ -59,7 +59,7 @@ _global_script_classes=[ {
 "language": "GDScript",
 "path": "res://src/main/ui/chat/chat-history.gd"
 }, {
-"base": "Sprite",
+"base": "Node2D",
 "class": "ChatIcon",
 "language": "GDScript",
 "path": "res://src/main/world/chat-icon.gd"
--- a/project/src/main/world/ChatIcon.tscn
+++ b/project/src/main/world/ChatIcon.tscn
@@ -9,7 +9,7 @@ length = 1.8
 loop = true
 step = 0.05
 tracks/0/type = "value"
-tracks/0/path = NodePath(".:scale")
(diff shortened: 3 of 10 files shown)


project/src/main/editor/creature/HookableNametag.tscn
```diff
@@ -1,8 +1,16 @@
...
+bg_color = Color( 0.188235, 0.188235, 0.376471, 1 )
+corner_radius_top_left = 25
+corner_radius_top_right = 25
+corner_radius_bottom_right = 25
+corner_radius_bottom_left = 25
 
 [node name="HookableNametag" type="Node2D"]
 
 [node name="Nametag" parent="." instance=ExtResource( 1 )]
+custom_styles/panel = SubResource( 1 )
```
Poobslag, line 16: Revert this; this class shouldn't have changed.

project/src/main/world/glowy-sphere.gd
```diff
@@ -5,3 +5,4 @@ Script for showing a placeholder chattable object in the overworld.
 
 func _ready() -> void:
 	set_meta("chat_path", "res://assets/main/dialog/glowy-sphere.json")
+	set_meta("chat_bubble_type", ChatIcon.THOUGHT)
```
Poobslag, line 8: This paradigm shift is worth documenting in the commit comment.

project/src/main/editor/creature/HookableNametag.tscn
Poobslag: Reverted.

project/src/main/world/glowy-sphere.gd
Poobslag: Expanded commit comment; `Creatures no longer have a chat icon -- instead, a separate ChatIcons node
generates chat icons for all creatures, and anything else that is in the
chattables group. Instead of customizing the chat icon by changing the
chat icon's properties, nodes can define metadata which affects their chat
icon appearance, such as the 'chat_bubble_type' metadata`

PR #593, opened 2020-09-08
Open tweak tab by default. Move mutagen slider.

Poobslag:
These creature editor changes match player's initial expectations a little
better. They expect to be able to pick creature parts. Locking/unlocking
features to change the creature generation is pretty esoteric, so it's OK
if that's hidden in a second tab.

Added hint about box colors.

Fixed bug where loading screen would stay up forever if the player
launched the CreatureEditor through the Godot editor, and then quit back
to the main menu.


project/src/main/editor/creature/reroll-control.gd
```diff
@@ -0,0 +1,33 @@
+extends Control
```
Poobslag, line 1: Maybe name this 'reroll-ui' for consistency with 'mutate-ui'

Poobslag: Renamed.

PR #623, opened 2020-09-17
Added cutscene logic for Ebe/Boatricia on overworld.

Poobslag:
Selecting certain scenarios from the level select screen now plays a
little cutscene.

The 'level select' chat tree now includes an extra 'chit chat' choice,
which is what they'll say on the overworld if they have no level.

launched_scenario_name -> launched_scenario_id. level_int -> level_num

Moved OverworldUi fields related to 'launched scenario', 'launched
creature dna' into Scenario.

### DIFF
--- a/project/src/main/world/overworld-world.gd
+++ b/project/src/main/world/overworld-world.gd
@@ -1,0 +1,38 @@
+extends Node
+"""
+Populates/unpopulates the creatures and obstacles on the overworld.
+"""
+
+export (NodePath) var creature_shadows_path: NodePath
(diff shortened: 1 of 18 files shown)


project/src/main/world/overworld-world.gd
```diff
@@ -0,0 +1,34 @@
+extends Node
```
Poobslag, line 1: Missing comments

Poobslag: Added

PR #625, opened 2020-09-17
Fixed issue where playfield lights were not always initialized properly.

### LINKED ISSUE #622: Playfield lights sometimes just randomly disappear
The playfield combo lights sometimes disappear and never come back. I've only seen this twice.

Tracing into the code, it appears that the `Playfield._color_tile_indexes` dictionary gets emptied. This dictionary is only modified on initialization, so this combined with the inconsistency of the bug makes it seem like a glitch in how Godot's garbage collector works. Maybe after 10-20 minutes the garbage collector incorrectly collects the dictionary because it thinks it isn't in use.

As a workaround, we could check if it's empty and reinitialize it, but this is sort of a scary thing to need to do.

Poobslag:
There was some logic which checked if the tileset was initialized, and if
it was already initialized, then the _color_tile_indexes map was left
empty. This logic was flawed because tilesets are sometimes cached if not
unique to a scene, but the _color_tile_indexes map needs to be initialized
every time.

These two objects are now initialized independently.

Closes #622.

### DIFF
--- a/project/src/main/puzzle/playfield-fx.gd
+++ b/project/src/main/puzzle/playfield-fx.gd
@@ -74,13 +74,14 @@ var _glow_duration := 0.0
 # the current background light color
 var _color := Color.transparent
 
-# tile indexes for each food/vegetable color
+# tile indexes by food/vegetable color
 var _color_tile_indexes: Dictionary
 
 func _ready() -> void:
-	reset()
 	PuzzleScore.connect("game_prepared", self, "_on_PuzzleScore_game_prepared")
 	_init_tile_set()
+	_init_color_tile_indexes()
+	reset()
 
 
 func _process(delta: float) -> void:
@@ -97,6 +98,9 @@ func reset() -> void:
 	_refresh_tile_maps()
 
 
+"""
+Initializes the different colored tiles in LightMap/GlowMap.
+"""
 func _init_tile_set() -> void:
 	if len($LightMap.tile_set.get_tiles_ids()) > 1:
 		return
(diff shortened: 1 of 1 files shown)


project/src/main/puzzle/playfield-fx.gd
```diff
@@ -112,10 +116,22 @@ func _init_tile_set() -> void:
...
+"""
+func _init_color_tile_indexes() -> void:
+	if _color_tile_indexes:
+		return
+	
+	for tile_index in $LightMap.tile_set.get_tiles_ids():
+		var color: Color = $LightMap.tile_set.tile_get_modulate(tile_index)
+		_color_tile_indexes[color] = tile_index
+
+
```
Poobslag, line 130: too much whitespace

project/src/main/puzzle/playfield-fx.gd
```diff
@@ -185,6 +201,10 @@ Calculates the new light pattern and refreshes the tile maps.
 func _refresh_tile_maps() -> void:
 	$BgStrobe.color = Utils.to_transparent(_color)
 	
+	if not _color_tile_indexes:
+		push_warning("PlayfieldFx._color_tile_indexes is empty, refreshing.")
```
Poobslag, line 205: Add a comment; we can probably remove this kludge eventually if this warning never happens.

Poobslag: I've removed the check; it seems overly paranoid. If the issue appears to recur, we can add something like this back in.

project/src/main/puzzle/playfield-fx.gd
Poobslag: Removed whitespace.

PR #631, opened 2020-09-20
Generic filler creatures get fatter over time.

Poobslag:
The filler characters weight isn't generated from scratch; it's now
pulled from a pool of 150 filler weights. If the player plays well these
weights will increase, if they play poorly their weights will decrease.

Extracted CREATURE_ID_PLAYER constant.

Moved CreatureEditor.FATNESSES constant into Global, so it can be reused
by CreatureLibrary.

### DIFF
--- a/project/src/main/ui/menu/main-menu.gd
+++ b/project/src/main/ui/menu/main-menu.gd
@@ -22,7 +22,7 @@ func _ready() -> void:
 
 func _launch_tutorial() -> void:
 	Global.clear_creature_queue()
-	var creature_def := CreatureLoader.load_creature_def_by_id("instructor")
+	var creature_def := CreatureLoader.load_creature_def("res://assets/main/creatures/instructor/creature.json")
 	Global.creature_queue.push_front(creature_def)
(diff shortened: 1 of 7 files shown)


project/src/main/ui/menu/main-menu.gd
```diff
@@ -22,7 +22,7 @@ func _ready() -> void:
 
 func _launch_tutorial() -> void:
 	Global.clear_creature_queue()
-	var creature_def := CreatureLoader.load_creature_def_by_id("instructor")
+	var creature_def := CreatureLoader.load_creature_def("res://assets/main/creatures/instructor/creature.json")
```
Poobslag, line 25: Worth documenting this fix.

todo.txt
```diff
@@ -0,0 +1,2 @@
+-player creature should still be normal
```

project/src/main/ui/menu/main-menu.gd
Poobslag: Documented, `Fixed incorrect instructor JSON path.`

todo.txt

PR #640, opened 2020-09-26
Added world locking/unlocking logic.

Poobslag:
Finishing all levels in world 1 unlocks world 2. 'hard locked' levels are now
hidden from the player. Each world has its own 'overall button' which displays
statistics specific to that world.

Added more placeholder levels, and made some earlier placeholder levels
easier. Moved wesker as primary creature, they're now in world 2.


project/src/main/ui/level-select/level-buttons.gd
```diff
@@ -29,19 +29,15 @@ var _columns := []
...
 var _column_width := COLUMN_WIDTH_SMALL
 
-# mapping from level IDs to LevelSettings instances
-var _level_settings_by_id: Dictionary
-
 var _duration_calculator := DurationCalculator.new()
 
-# tracks which levels unlock other levels
-var _world_locks := WorldLocks.new()
+# keeps tracks of which worlds/levels unlock other worlds/levels.
```
Poobslag, line 34: grammar; 'keeps tracks'

project/src/main/ui/level-select/world-locks.gd
```diff
@@ -1,142 +0,0 @@
-class_name WorldLocks
```
Poobslag, line 1: Worth documenting that this became 'LevelSelectModel'

project/src/main/ui/level-select/level-buttons.gd
Poobslag: Changed, `keeps track of which worlds...`

project/src/main/ui/level-select/world-locks.gd
Poobslag: Improved commit comment; `Renamed WorldLocks to LevelSelectModel.`

PR #662, opened 2020-10-01
Added buttons for talk/settings/phone on overworld.

Poobslag:
Level select screen is now available via 'phone' overworld button, instead
of on the main menu. Hopefully this sort of feels like the 'map' button in games
like Diablo and Among Us. Thematically, you'll be calling someone on your cell
phone, but mechanically, I want it to feel like a map or a level select.

The 'talk' and 'settings' overworld buttons are now always visible,
even when not using a touchscreen.

The level select's 'world buttons' are white, to help them stand out.
(Eventually they'll have pictures and stand out more.)


project/project.godot
```diff
@@ -772,6 +772,14 @@ ui_cancel={
...
  ]
 }
+ui_focus_next={
+"deadzone": 0.5,
+"events": [  ]
+}
+ui_focus_prev={
+"deadzone": 0.5,
+"events": [  ]
+}
```
Poobslag, line 782: Document this `focus_next`, `focus_prev` change

project/src/main/ui/level-select/LevelSelectButton.tscn
```diff
@@ -41,13 +41,13 @@ margin_left = 4.0
...
 margin_bottom = -4.0
-size_flags_horizontal = 0
-size_flags_vertical = 0
+size_flags_vertical = 1
 custom_fonts/font = ExtResource( 6 )
 text = "Just getting started"
 align = 1
 valign = 1
 autowrap = true
+clip_text = true
```
Poobslag, line 50: Maybe good to mention this `clip_text` change in the commit comment, since it was tricky.

project/project.godot
Poobslag: Expanded commit comment, `Removed 'focus_next' and 'focus_prev' input events `

project/src/main/ui/level-select/LevelSelectButton.tscn
Poobslag: Expanded commit comment, `Enabled 'clip_text' property for level select buttons `

PR #673, opened 2020-10-10
Improved tutorial chat window appearance. Minor chat/tutorial fixes.

Poobslag:
Other chat windows in the game had a background texture which smooshes and
grows. Now the tutorial windows use the same style.

Fixed bug where chat frames were only growing on the right/bottom The chat
dialog frames were controls, which expand from the upper-left corner unless a
pivot is defined.

Fixed bug where visual box clears were't incrementing enough. If you
cleared 3 lines with one piece, it would still continue the tutorial but
the box clears indicator would only increment by 1.


project/src/main/world/creature/creature-loader.gd
```diff
@@ -8,6 +8,8 @@ class converts that information into granular information such as 'her Eye/Sprin
 loads resource files specific to each creature.
 """
 
+const INSTRUCTOR_PATH := "res://assets/main/creatures/instructor/creature.json"
```
Poobslag, line 11: comment

project/src/main/ui/chat/chat-line-panel.gd
```diff
@@ -13,12 +13,7 @@ const PULSE_AMOUNT := Vector2(0.015, 0.030)
 const CHAT_TEXTURE_COUNT := 16
 
 # Longer lines of dialog are displayed in bigger panels
```
Poobslag, line 15: This comment is slightly out of date, and should be updated

project/src/main/world/creature/creature-loader.gd
Poobslag: Added comment, `the creature definition path for the instructor who leads tutorials`

project/src/main/ui/chat/chat-line-panel.gd
Poobslag: Improved comment, `Different panel sizes to try, ordered from smallest to largest.`

PR #682, opened 2020-10-15
Scenarios can now impose a piece limit.

Poobslag:
New levels, 'veggie patty', 'accelerator'.

We now perform 'speed ups' before line clears, because it sounds better
for speedups caused by a piece milestone. Otherwise the sound effects play when
the player drops a piece, except when they clear lines, and then there's a
delay, which is jarring.

We still perform 'level finishes' after line clears, because otherwise the
player won't earn the points from their final line clear.

For this reason MilestoneManager now performs most level checks after line
clears, instead of checking for them every frame. This also required some
minor tweaks to signals in LineClearer and PuzzleScore.

For levels with limited pieces, the next piece queue runs out.

### DIFF
--- a/project/src/main/puzzle/piece/piece-queue.gd
+++ b/project/src/main/puzzle/piece/piece-queue.gd
@@ -6,13 +6,17 @@ Queue of upcoming randomized pieces.
 This queue stores the upcoming pieces so they can be displayed, and randomizes them according to some complex rules.
 """
 
-# minimum number of next _pieces in the queue, before we add more
+# minimum number of next pieces in the queue, before we add more
 const MIN_SIZE := 50
+
+const UNLIMITED_PIECES := 999999
 
 var _pieces := []
 
 # default pieces to pull from if none are provided by the level
 var _default_piece_types := PieceTypes.all_types
+
+var _remaining_piece_count := UNLIMITED_PIECES
 
 func _ready() -> void:
 	Level.connect("settings_changed", self, "_on_Level_settings_changed")
@@ -24,6 +28,10 @@ func _ready() -> void:
 Clears the pieces and refills the piece queues.
 """
 func clear() -> void:
+	if Level.settings.finish_condition.type == Milestone.PIECES:
+		_remaining_piece_count = Level.settings.finish_condition.value
+	else:
+		_remaining_piece_count = UNLIMITED_PIECES
 	_pieces.clear()
 	_fill()
 
@@ -33,6 +41,8 @@ Pops the next piece off the queue.
 """
 func pop_next_piece() -> PieceType:
 	var next_piece_type: PieceType = _pieces.pop_front()
+	if _remaining_piece_count != UNLIMITED_PIECES:
+		_remaining_piece_count -= 1
 	_fill()
 	return next_piece_type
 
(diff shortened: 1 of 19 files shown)


project/src/main/puzzle/rank-calculator.gd
```diff
@@ -127,13 +128,17 @@ func _max_lpm() -> float:
...
+					level_lines = speed_up.value / 2
 				Milestone.TIME_OVER:
 					level_lines = speed_up.value * 60 / frames_per_line
 				Milestone.SCORE:
 					level_lines = speed_up.value / \
 							(master_box_score(Level.settings) + master_combo_score(Level.settings) + 1)
 		elif finish_condition.type == Milestone.LINES:
 			level_lines = finish_condition.value
+		elif finish_condition.type == Milestone.PIECES:
+			level_lines = finish_condition.value / 2
```
Poobslag, line 141: Suppress integer division warnings.

project/src/main/puzzle/rank-result.gd
```diff
@@ -21,6 +21,10 @@ var speed_rank := WORST_RANK
 var lines := 0
 var lines_rank := WORST_RANK
 
+# number of pieces placed
+var pieces := 0
+var pieces_rank := WORST_RANK
```
Poobslag, line 26: These fields need to be populated/saved by `to_json_dict`, `from_json_dict`

project/src/main/puzzle/tutorial/tutorial-module.gd
```diff
@@ -15,7 +15,7 @@ var playfield: Playfield
 var piece_manager: PieceManager
 
 func _ready() -> void:
-	puzzle = hud.puzzle
+	puzzle = hud.puzzlet
```
Poobslag, line 18: puzzlet?

project/src/main/puzzle/piece/piece-queue.gd
```diff
@@ -9,11 +9,15 @@ This queue stores the upcoming pieces so they can be displayed, and randomizes t
...
 const MIN_SIZE := 50
 
+const UNLIMITED_PIECES := 999999
+
 var _pieces := []
 
 # default pieces to pull from if none are provided by the level
 var _default_piece_types := PieceTypes.all_types
 
+var _remaining_piece_count := UNLIMITED_PIECES
```
Poobslag, line 19: This value is never decremented, so it will not work for larger values.

project/src/main/puzzle/rank-calculator.gd
Poobslag: Suppressed.

project/src/main/puzzle/rank-result.gd
Poobslag: Fields are now converted to/from json.

project/src/main/puzzle/tutorial/tutorial-module.gd
Poobslag: Removed.

project/src/main/puzzle/piece/piece-queue.gd
Poobslag: Value is now decremented properly. I tested it with a piece queue size of 20, and verified it was broken before, and works OK now.

PR #686, opened 2020-10-20
Added jitter fix. Split DnaLoader from CreatureVisuals.

Poobslag:
Jitter seems to mostly stem from editing too many shader parameters in one
frame; this is now spread out over multiple frames.

DnaLoader is now separate because it needs to have pausing disabled; otherwise
creature colors don't update properly when using the color pickers in the
creature editor.

Note: I initially attempted an approach where instead of loading shader
parameters gradually, I loaded all DNA properties gradually. This
introduced problems because the code would try to animate partially-loaded
textures, and required stuff like mutexes and new signals to keep track of
whether the DNA was unloaded, partially loaded, or fully loaded. It was a
big mess, but if you want to pursue this in the future you can see my work
in 59e1f6e.

### DIFF
--- a/delint.sh
+++ b/delint.sh
@@ -19,8 +19,8 @@ grep -R -n "var [^:]* = " --include="*.gd" ./project/src \
   | grep -v "chat-event.gd:74" \
   | grep -v "level-settings.gd:182" \
   | grep -v "level-settings.gd:184" \
-  | grep -v "creature-visuals.gd.*var property_value =" \
-  | grep -v "creature-visuals.gd.*var shader_value ="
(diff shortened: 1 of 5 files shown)


project/project.godot
```diff
@@ -719,7 +719,7 @@ _global_script_class_icons={
 config/name="Turbo Fat"
 run/main_scene="res://src/main/ui/menu/LoadingScreen.tscn"
 config/icon="res://assets/main/ui/icon.png"
-config/version="0.1011"
+config/version="0.1011b"
```
Poobslag, line 722: Revert this.

Poobslag: Reverted

PR #688, opened 2020-10-22
Reduced jitter by improving PackedSprite performance with caching.

Poobslag:
PackedSprite parses JSON files which specify the regions for each frame.
This takes about 1-2 milliseconds, which adds up to about 20 milliseconds
when loading a new creature.

These rectangles are now cached in ResourceCache, which accelerates
CreatureVisuals._load_dna() from 20 ms to about 1 ms. This sharply reduces
the jitter when new creatures are loaded during puzzles.

### DIFF
--- a/project/src/main/resource-cache.gd
+++ b/project/src/main/resource-cache.gd
@@ -27,8 +27,17 @@ export (bool) var verbose := false
 export (bool) var minimal_resources := false
 
 # maintains references to all resources to prevent them from being cleaned up
+# key: resource path
+# value: resource
 var _cache := {}
 var _cache_mutex := Mutex.new()
+
+# stores parsed versions of Aseprite json resources to speed up retrieval
+# key: json resource path
+# value: array of Rect2 instances representing regions defined by Aseprite
+var _frame_src_rect_cache := {}
+var _frame_dest_rect_cache := {}
+var _frame_cache_mutex := Mutex.new()
 
 # setting this to 'true' causes the background thread to terminate gracefully
 var _exiting := false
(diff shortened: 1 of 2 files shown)


project/src/main/resource-cache.gd
```diff
@@ -27,9 +27,18 @@ export (bool) var verbose := false
 export (bool) var minimal_resources := false
 
 # maintains references to all resources to prevent them from being cleaned up
+# key: resource path
+# value: resource
 var _cache := {}
 var _cache_mutex := Mutex.new()
 
+# stores parsed version of json resources to speed up retrieval
```
Poobslag, line 35: Comment should probably mention something about "Aseprite" or "regions". This isn't generic enough to apply to all json resources.

project/src/main/resource-cache.gd
```diff
@@ -203,6 +233,8 @@ func _find_resource_paths() -> Array:
 				_remaining_scene_paths.append("%s/%s" % [dir.get_current_dir(), file.get_file()])
 			elif file.ends_with(".png.import") or file.ends_with(".wav.import"):
 				_remaining_resource_paths.append("%s/%s" % [dir.get_current_dir(), file.get_basename()])
+			elif file.ends_with(".json"):
+				_remaining_resource_paths.append("%s/%s" % [dir.get_current_dir(), file.get_file()])
```
Poobslag, line 237: This could be lumped in with the `.tscn` case above.

project/src/main/resource-cache.gd
```diff
@@ -229,6 +261,34 @@ func _find_resource_paths() -> Array:
 	return _remaining_resource_paths
 
 
+func _load_json_resource(json_path: String) -> void:
```
Poobslag, line 264: Add a method-level comment.

project/src/main/resource-cache.gd
```diff
@@ -252,3 +312,7 @@ func _load_resource(resource_path: String) -> void:
 		_cache_mutex.lock()
 		_cache[resource_path] = result
 		_cache_mutex.unlock()
+
+
+static func _json_to_rect2(json: Dictionary) -> Rect2:
```
Poobslag, line 317: Add a method-level comment.

project/src/main/resource-cache.gd
Poobslag: The two cases are slightly different; they cannot be combined.

project/src/main/resource-cache.gd
Poobslag: Added.

project/src/main/resource-cache.gd
Poobslag: Added.

project/src/main/resource-cache.gd
Poobslag: Changed to `...versions of Aseprite json resources...`

PR #697, opened 2020-10-31
Added combo tutorial. Tutorial improvements, such as replays.

### LINKED ISSUE #694: Squish move tutorial should use a computer-controlled piece
The squish move tutorial has a part where the instructor explains "you can use squish moves to recover from a mistake". But, since there's currently no way for the instructor to steer a piece around, those parts just set up the level with the mistake already made. It would be better to see the instructor make a mistake.

-add logic for recording/replaying input
 -recording saves a series of int bitmasks for player input
 -replaying replays the saved input
 -replaying ignores actual player's input
 -bitmasks are compressed (PoolByteArray.compress?)
-steer piece in part 5 to a dumb place
-steer piece in part 6 to a dumb place

Poobslag:
Pulled 'hide_completed_skill_tally_items' method into TutorialModule
class.

Combo resets based on line clears, not score. This is to accommodate
tutorials where score is not tracked, but line clears are.

Added 'ordered_start', 'enhance_combo_fx', 'input_replay' level settings.
These are needed to support the combo tutorial, although 'ordered_start'
could perhaps be used to reduce (or eliminate) randomness in a particular
level.

We now hide skill tally items after level change, instead of before the
level change. They were disappearing too soon before, it's nice to leave
them onscreen so the player can see what they've accomplished.

Changed tutorial 'game_prepared' signal timing. Tutorials often want to set up
things like the number of cleared lines or the current combo state, but this
fights with other 'game_prepared' signals like the ComboTracker which want to do
the same thing. By changing this to an 'after_game_prepared' signal, we ensure
the Tutorial's special case logic overrides the default logic.

Fixed bug where tutorial diagrams would block the "start/quit" UI box, making it
so you couldn't quit during certain parts of the tutorial.

The tutorial sections the player's attempted are now cleared when
restarting a tutorial. This makes it so the player isn't told "Try again!"
when restarting a tutorial from the beginning.

Squish tutorials now move pieces around.

The squish move tutorial now uses a computer-controlled piece. This helps
convey the idea of misplacing a piece and cleaning up afterwards.

Closes #694.


project/src/main/puzzle/playfield-fx.gd
```diff
@@ -176,16 +180,21 @@ Calculates the brightness of the lights based on the current combo.
...
+	if Level.settings.other.enhance_combo_fx and combo >= 3:
+		combo += 12
+	
 	if combo >= 12:
 		# 100% opacity flashing lights is hard on the eyes. even at max combo, we dial it back from 100%
 		_brightness = lerp(0.30, 0.60, smoothstep(12, 40, combo))
 		_pulse_amount = lerp(0.10, 0.30, smoothstep(12, 30, combo))
 		_glow_duration = lerp(0.60, 2.40, smoothstep(12, 100, combo))
-	elif combo >= 5:
+	elif combo >= 3:
```
Poobslag, line 193: Mention this change in the commit comment.

project/src/main/puzzle/input-replay.gd
```diff
@@ -0,0 +1,70 @@
...
+
+
+"""
+Returns 'true' if the replay data specifies that the action was released on the current input frame.
+
+This only true for the exact frame when the release is first triggered.
+"""
+func is_action_released(action: String) -> bool:
+	var action_frame_key := "%s -%s" % [PuzzleScore.input_frame, action]
+	var _released := _action_timings.has(action_frame_key)
```
Poobslag, line 51: Change to `released`; we don't prefix local variables with underscores.

project/src/main/puzzle/input-replay.gd
```diff
@@ -0,0 +1,70 @@
...
+
+
+"""
+Returns 'true' if the replay data specifies that the action was pressed on the current input frame.
+
+This only true for the exact frame when the press is first triggered.
+"""
+func is_action_pressed(action: String) -> bool:
+	var action_frame_key := "%s +%s" % [PuzzleScore.input_frame, action]
+	var _pressed := _action_timings.has(action_frame_key)
```
Poobslag, line 38: Change to `released`; we don't prefix local variables with underscores.

project/src/main/puzzle/piece/frame-input.gd
```diff
@@ -13,36 +13,48 @@ export (String) var action: String
...
 export (String) var cancel_action: String
 
+var pressed_frames: int
+
 var _just_pressed: bool
 var _pressed: bool
-
-var frames: int
 var _buffer: bool
+var _print_inputs: bool = false
```
Poobslag, line 21: Remove redundant `= false`

project/src/main/puzzle/piece/frame-input.gd
```diff
@@ -13,36 +13,48 @@ export (String) var action: String
...
 
 func _unhandled_input(event: InputEvent) -> void:
+	if not Level.settings.input_replay.empty():
+		# don't process player input during replays
```
Poobslag, line 25: Less ambiguously, `don't process button presses when replaying prerecorded input`

project/src/main/puzzle/playfield-fx.gd
Poobslag: New commit comment, `Combo background effects now start at combo level 3`

project/src/main/puzzle/input-replay.gd
Poobslag: Changed

project/src/main/puzzle/input-replay.gd
Poobslag: Changed

project/src/main/puzzle/piece/frame-input.gd
Poobslag: Changed to `var _print_inputs := false`; this makes it slightly more intuitive to toggle it between true and false, while still avoiding the redundant `bool = false`

project/src/main/puzzle/piece/frame-input.gd
Poobslag: Changed

PR #705, opened 2020-11-03
Refactored CreatureDef.min_fatness and breadcrumb pop logic

Poobslag:
CreatureDef now defines min_fatness, not fatness. The CreatureDef json still
uses 'fatness' for simplicity and backwards compatibility.

Moved duplicated 'popping top level of breadcrumb trail' logic into
Breadcrumb class. This unfortunately makes Breadcrumb a little more
project-specific, since it depends on ResourceCache and LoadingScreen.

### DIFF
--- a/project/src/main/editor/creature/creature-editor.gd
+++ b/project/src/main/editor/creature/creature-editor.gd
@@ -48,8 +48,6 @@ func _ready() -> void:
 	_name_generator = NameGenerator.new()
 	_name_generator.load_american_animals()
 	
-	Breadcrumb.connect("trail_popped", self, "_on_Breadcrumb_trail_popped")
-	
 	for allele in DnaUtils.BODY_PART_ALLELES:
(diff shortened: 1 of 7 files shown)


project/src/main/editor/creature/creature-editor.gd
```diff
@@ -386,11 +388,6 @@ func _random_highlight_color(color: Color) -> Color:
 	return new_color
 
 
-func _on_Breadcrumb_trail_popped(_prev_path: String) -> void:
-	if not Breadcrumb.trail:
-		get_tree().change_scene("res://src/main/ui/menu/LoadingScreen.tscn")
-
-
 func _on_Quit_pressed() -> void:
```
Poobslag, line 391: With this function removed, you can disconnect the signal to avoid an `Error calling method from signal 'trail_popped'` error

Poobslag: Disconnected signal

PR #709, opened 2020-11-04
Refactored combos, restaurant nametags, dna loader.

Poobslag:
Combined ComboTracker.combo and PuzzleScore.creature_line_clears. We were
storing this information in two places before. It's now stored in PuzzleScore's
new combo field.

Pushed logic for refreshing nametag appearance into NametagPanel to keep
RestaurantView smaller.

DnaLoader no longer emits a signal on CreatureVisuals. Nodes should emit
their own signals, even if it's technically a little less efficient in
this case.

### DIFF
--- a/project/src/main/puzzle/combo-tracker.gd
+++ b/project/src/main/puzzle/combo-tracker.gd
@@ -6,14 +6,10 @@ Keeps track of the player's combo.
 Increments the combo when the player does well, breaks the combo when the player messes up.
 """
 
-signal combo_changed(value)
 signal combo_break_changed(value)
 
 # bonus points which are awarded as the player continues a combo
 const COMBO_SCORE_ARR = [0, 0, 5, 5, 10, 10, 15, 15, 20]
-
-# number of lines the player has cleared without dropping their combo
-var combo := 0 setget set_combo
 
 # The number of pieces the player has dropped without clearing a line or making a box.
 var combo_break := 0 setget set_combo_break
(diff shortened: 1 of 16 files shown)


project/src/main/puzzle/combo-tracker.gd
```diff
@@ -6,15 +6,11 @@ Keeps track of the player's combo.
 Increments the combo when the player does well, breaks the combo when the player messes up.
 """
 
-signal combo_changed(value)
 signal combo_break_changed(value)
 
 # bonus points which are awarded as the player continues a combo
 const COMBO_SCORE_ARR = [0, 0, 5, 5, 10, 10, 15, 15, 20]
 
-# number of lines the player has cleared without dropping their combo
```
Poobslag, line 15: Preserve this comment; the comment in PuzzleScore.combo is less precise

project/src/main/puzzle/puzzle-score.gd
```diff
@@ -276,12 +276,13 @@ func get_creature_score() -> int:
 	return creature_scores[creature_scores.size() - 1]
 
 
-func get_creature_line_clears() -> int:
-	return creature_line_clears[creature_line_clears.size() - 1]
-
-
-func set_creature_line_clears(value: int) -> void:
-	creature_line_clears[creature_line_clears.size() - 1] = value
+func set_combo(value: int) -> void:
```
Poobslag, line 279: Our guidelines specify that setters follow the convention `set_foo(new_foo: Type) -> void`

project/src/main/puzzle/restaurant-view.gd
```diff
@@ -118,7 +111,11 @@ func _on_PuzzleScore_game_prepared() -> void:
 		scroll_to_new_creature()
 
 
-func _on_PuzzleScore_combo_ended() -> void:
+func _on_PuzzleScore_combo_changed(value: int) -> void:
+	if value > 0:
+		# this method only cares about the combo resetting to 0
```
Poobslag, line 116: Rephrase this comment to be slightly more formal

project/src/main/puzzle/combo-tracker.gd
Poobslag: PuzzleScore.combo description has been updated, ` # The number of lines the player has cleared without dropping their combo`

project/src/main/puzzle/puzzle-score.gd
Poobslag: Changed to `func set_combo(new_combo: int) -> void:`

project/src/main/puzzle/restaurant-view.gd
Poobslag: Changed to `if the combo is not resetting, we ignore the change`

PR #710, opened 2020-11-05
Removed redundant Playfield.after_piece_written signal

Poobslag:
Tutorial combos are now cleared after the level changes, instead of before
the level changes. That way the background stays lit up.


project/src/main/puzzle/puzzle-score.gd
```diff
@@ -211,15 +211,17 @@ func add_line_score(combo_score: int, box_score: int) -> void:
 	_add_creature_score(1 + combo_score + box_score)
 	_add_bonus_score(combo_score + box_score)
 	_add_score(1)
-
+	
 	emit_signal("added_line_score", combo_score, box_score)
+	emit_signal("combo_changed", combo)
```
Poobslag, line 216: For consistency, we should continue emitting `combo_changed` after `score_changed`

Poobslag: Moved.

PR #711, opened 2020-11-06
Rewrote piece shuffling algorithm to be more reliable.

Poobslag:
The old algorithm shuffled pieces up to 48 times and hoped that one of the
random arrangements resulted in having no back-to-back pieces. But it was
random, and even though the most I ever saw it take was 15-20 tries, it
could theoretically have bad luck and fail to come up with an arrangement.

The new algorithm finds a location for each piece deterministically, and
is more reliable. It will only fail if it's impossible to find an
arrangement (e.g a custom level with a piece queue with 3 T pieces and an
O piece)

### DIFF
--- a/project/src/main/puzzle/piece/piece-queue.gd
+++ b/project/src/main/puzzle/piece/piece-queue.gd
@@ -146,19 +146,49 @@ func _fill_remaining_pieces() -> void:
 	while _pieces.size() < MIN_SIZE:
 		# fill a bag with one of each piece and one extra; draw them out in a random order
 		var new_pieces := shuffled_piece_types()
-		
-		# avoid having two of the same piece consecutively
-		if new_pieces.size() >= 3:
-			for _i in range(48):
-				if _has_duplicate_pieces(new_pieces):
-					new_pieces.shuffle()
-				elif not _pieces.empty() and _pieces.back() == new_pieces[0]:
-					new_pieces.pop_front()
-					new_pieces.insert(int(rand_range(1, new_pieces.size() + 1)), _pieces.back())
-				else:
-					break
 		_pieces += new_pieces
+		
+		if new_pieces.size() >= 3:
+			# for levels with multiple identical pieces in the bag, we shuffle the bag so that those identical pieces
+			# aren't back to back
+			var min_to_index := _pieces.size() - new_pieces.size()
+			var from_index := min_to_index
+			while from_index < _pieces.size():
+				var to_index := from_index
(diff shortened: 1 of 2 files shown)


project/src/test/puzzle/piece/test-piece-queue.gd
```diff
@@ -0,0 +1,24 @@
+extends "res://addons/gut/test.gd"
+"""
+Unit test demonstrating duration calculation logic.
```
Poobslag, line 3: This comment is inaccurate.

project/src/main/puzzle/piece/piece-queue.gd
```diff
@@ -146,18 +146,31 @@ func _fill_remaining_pieces() -> void:
...
-					new_pieces.insert(int(rand_range(1, new_pieces.size() + 1)), _pieces.back())
-				else:
-					break
-		_pieces += new_pieces
+			# loop over each of the newly added pieces
+			var min_piece_index := _pieces.size() - new_pieces.size()
+			var piece_index := min_piece_index
+			while piece_index < _pieces.size():
+				if _pieces[piece_index] == _pieces[piece_index - 1]:
+					# a piece appears back-to-back in the queue. find a new place for it
```
Poobslag, line 158: For documentation purposes (and possibly to get away from the pyramid-of-doom here) it might make sense to extract this in a new function; maybe something like `move_duplicate_piece`

Poobslag: It would need to return something like `new_position` in the response so that we can futz with `piece_index` appropriately

Poobslag: Try it out -- I'm mostly concerned with avoiding having 5-6 levels of indentation as becomes difficult to read and opens the door for bugs (e.g incrementing piece_index conditionally instead of on every iteration of a loop)

Alternatively, we could pull out the entire loop as something like `fix_duplicate_pieces` or something like that.

project/src/test/puzzle/piece/test-piece-queue.gd
Poobslag: Fixed comment; `Unit test for the queue of upcoming pieces.`

project/src/main/puzzle/piece/piece-queue.gd
Poobslag: Extracted new function, `func _move_duplicate_piece(from_index: int, min_to_index: int) -> int:`. It wasn't as ugly as I thought.

PR #713, opened 2020-11-07
Levels are arranged hierarchically.

### LINKED ISSUE #235: Store scenarios hierarchically
Currently all of our scenarios are in a big folder with names like `boatricia.json` for boatricia's scenarios, and `tutorial-beginner-0.json` for tutorial scenarios. This is not maintainable. Eventually someone could create a character named `tutorial` and wreak havoc with our naming scheme.

Scenarios should be stored hierarchically:

```
/scenario/
/scenario/customer/boatricia/boatricia.json
/scenario/tutorial/beginner/0.json
/scenario/practice/marathon-expert.json
```

Poobslag:
Added backwards compatibility logic to preserve tutorial/score data
despite renames.

Closes #235.

Removed incorrectly written backwards-compatibility test; this was
intended to demonstrate backwards compatibility with a specific save
version, but no save file was being loaded, so it was just reading the
player's current save file.


todo.txt
```diff
@@ -0,0 +1,5 @@
+*move levels around
+*ensure scores still save
```
Poobslag, line 2: remove this

PR #727, opened 2020-11-11
Fixed bug where survival level ended prematurely.

Poobslag:
Survival mode was ending immediately after the last line clear, instead of
waiting for the end-of-level 'big finish' clears.

This specific glitch was introduced in
1a230842, but I believe the genesis for these bugs was
cab934b0d3 which replaced a 'combo_ended' signal with a 'combo_changed'
signal. These are interchangeable sometimes but there are scenarios where
a combo can end without the value being reset to zero, such as at the end
of a 'customer limit' level.

Added new 'micro-' levels for quickly testing end-of-level glitches.

Moved test levels into main/puzzle/levels to mirror test/puzzle/levels


project/src/main/editor/puzzle/level-editor.gd
```diff
@@ -7,7 +7,7 @@ Full instructions are available at https://github.com/Poobslag/turbofat/wiki/lev
 """
 
 # default to an empty level; players may be confused if it's not empty
-const DEFAULT_LEVEL := "ultra_normal"
+const DEFAULT_LEVEL := "practice/ultra_normal"
```
Poobslag, line 10: Worth adding to commit comment.

Poobslag: Added.

PR #735, opened 2020-11-29
Fixed bug where all chat icons blinked sometimes.

Poobslag:
When the player runs near different creatures, their chat icon is supposed
to light up. But when this happened, all of the chat bubbles blinked. This
is because the tween's parameters were using the wrong node's modulate
proeprty.

I also added a check to avoid launching unnecessary chat icon tweens. This
shouldn't have any visible effect, but fewer tweens is good for performance and
simplicity.

### DIFF
--- a/project/src/main/world/chat-icon.gd
+++ b/project/src/main/world/chat-icon.gd
@@ -67,6 +67,9 @@ Make the chat icon focused, so the player knows they can interact with it.
 This brightens the icon and makes it bounce faster.
 """
 func focus() -> void:
+	if _focused == true:
+		return
+	
 	_tween_modulate(COLOR_FOCUSED, 0.8)
 	_focused = true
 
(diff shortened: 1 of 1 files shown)


project/src/main/world/Overworld.tscn
```diff
@@ -83,13 +83,13 @@ viewport_path = NodePath("World/Ground/GoopViewport")
...
+shader_param/black = Color( 0.254902, 0.156863, 0.117647, 1 )
 shader_param/red_texture_scale = Vector2( 1, 1 )
-shader_param/red_texture_offset = Vector2( 0, 0 )
 shader_param/green_texture_scale = Vector2( 1, 1 )
-shader_param/green_texture_offset = Vector2( 0, 0 )
 shader_param/blue_texture_scale = Vector2( 1, 1 )
+shader_param/red_texture_offset = Vector2( 0, 0 )
+shader_param/green_texture_offset = Vector2( 0, 0 )
 shader_param/blue_texture_offset = Vector2( 0, 0 )
-shader_param/black = Color( 0.254902, 0.156863, 0.117647, 1 )
```
Poobslag, line 92: Roll this back.

Poobslag: Reverted

PR #754, opened 2020-12-09
Extracted LevelLibrary. Fixed possible 'inaccessible tutorial' bug.

Poobslag:
Other classes and logic need to know things like 'which levels does the
player have unlocked' outside of the level select UI, and it makes sense
for this to be a singleton node so we don't need to reload the JSON
everywhere.

Fixed potential tutorial bug. There was some logic setting the tutorial world to
LevelLock.STATUS_CLEARED (1) if all levels in the tutorial were complete.
However, tutorial worlds don't use this particular enum, and it resulted in the
tutorials being marked as LockStatus.LOCK (1) instead.

### DIFF
--- a/project/src/main/player-save.gd
+++ b/project/src/main/player-save.gd
@@ -148,6 +148,9 @@ func _load_player_data_from_file(filename: String) -> bool:
 		save_item.from_json_dict(json_save_item_obj)
 		_load_line(save_item.type, save_item.key, save_item.value)
 	
+	# emit a signal indicating the level history was loaded
+	PlayerData.emit_signal("level_history_changed")
+	
 	return true
 
 
--- a/project/project.godot
+++ b/project/project.godot
@@ -259,6 +259,11 @@ _global_script_classes=[ {
 "language": "GDScript",
 "path": "res://src/main/puzzle/leaf-poof.gd"
 }, {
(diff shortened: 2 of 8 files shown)


project/src/main/player-save.gd
```diff
@@ -148,9 +148,15 @@ func _load_player_data_from_file(filename: String) -> bool:
 		save_item.from_json_dict(json_save_item_obj)
 		_load_line(save_item.type, save_item.key, save_item.value)
 	
+	# emit a signal indicating the level history was loaded
+	PlayerData.emit_signal("level_history_changed")
+	
 	return true
 
 
+var something := 0
```
Poobslag, line 157: remove

project/src/main/ui/chat/chat-library.gd
```diff
@@ -6,6 +6,7 @@ Dialog is stored as a set of json resources. This class parses those json resour
 into the UI.
 """
 
+
```
Poobslag, line 9: revert

Poobslag: Reverted

project/src/main/player-save.gd
Poobslag: Removed.

PR #755, opened 2020-12-10
Eliminated chat menus, locked levels. Cache filler chats.

### LINKED ISSUE #642: Overworld creatures shouldn't offer locked levels.
If levels are locked, the overworld creatures shouldn't let you play them.

- Talking to people on overworld automatically picks their lowest unbeaten, unlocked level, or some filler dialog if there is none
- Overworld people have a different icon for "play a level with me"

This can also end the whole '(level 1) (level 2) (chit chat)' submenu, so that code should go away.

- remove (level 1) (level 2) dialog trees
- remove code referencing those dialog trees
- utilize chat_tree_cache after removing dialog trees; including for filler dialog

Poobslag:
Overworld creatures like Ebe and Boatricia no longer offer 'level-1',
'level-2', 'chit-chat' options. They now always pick the oldest unbeaten
level when you speak to them, and dialog if no levels are available.

As a result, overworld creatures will no longer let you pick locked levels
by talking to them either.

All overworld dialog is now cached to ensure the player doesn't cycle
through all of a creature's dialog quickly. ...This cache mechanism was in
place before, but it only activated when the player selected one of the
choices from the chat menus.

Closes #642.


Poobslag:
I first attempted to complete this a few months ago, but took a second attempt this week. My first attempt conceptualized a different 'chat flow' which is worth documenting because it's interesting, and because i might eventually decide the approach I went with is too simple:

```
-overworld picks oldest level
 -pick level 1; no popup
 -pick least recently played unlocked level
 -if tied, pick lowest alphabetically
 -if there are no unlocked levels, it picks chit-chat
-overworld sometimes picks chit-chat
 -50% of all creatures do 'chit chat' instead
 -cycles when a level is cleared
```

project/src/main/ui/chat/chat-library.gd
```diff
@@ -18,36 +18,36 @@ Parameters:
...
+		chat_tree.from_json_dict(creature.dialog[chosen_dialog])
+		chat_tree.history_key = "dialog/%s/%s" % [creature.creature_id, chosen_dialog]
 	else:
-		# returning a tree where the player can select creature dialog
+		var path := "res://assets/main/creatures/primary/%s/%s.json" % \
+				[creature.creature_id, chosen_dialog.replace("_", "-")]
+		chat_tree = load_chat_events_from_file(path)
+	
+	# set the launched_level property, so that levels will launch after the dialog completes
+	if state["level_num"] != -1:
```
Poobslag, line 46: The old logic had `level_num >= 1` which is a little more bulletproof

Poobslag: Changed to `>= 1`. Testing this code also helped me catch another bug which I fixed, where some code was referencing `level_num` instead of `state["level_num"]`.

PR #760, opened 2020-12-20
Preliminary version of indoor overworld scene.

Poobslag:
The player will eventually be able to walk inside and outside, and there
will be cinematic scenes which take place indoors.

OverworldIndoors.tscn isn't accessible from anywhere, but it can be
launched independently. The player can move, and there are
stools/tables/creatures.

obstacle-map-shadows.gd can now render shadows for tiles which span
multiple cells. This is necessary for the interior scenes, where there are
walls which are three cells wide.

Extracted goop-ground.gd from overworld-world.gd.


todo.txt
```diff
@@ -0,0 +1,88 @@
...
+-can go indoors
+-can go outdoors
+-some characters spawn indoors
+-some characters spawn outdoors
+-first indoor story character always spawns indoors if their cutscene demands it
+
+
+
+ideas for handling wide shadows?
+
```
Poobslag, line 87: Remove this file.

PR #790, opened 2021-01-25
Initial cutscene location logic. Level dialog is no longer on creatures.

Poobslag:
Cutscenes can take place indoors or outdoors. Added logic for spawning
cutscene characters at different locations.

Level dialog used to be attached to creatures who had specific logic for
'use this dialog for level 1, this dialog for level 2.' This logic is now
implicit, and level dialog is in /assets/main/puzzle/levels/cutscenes

Levels no longer have 'level numbers'. A creature used to have the idea of
their first and second level. This is now implicit based on the contents
of worlds.json and how the levels are organized.

Creature clipping is disabled during cutscene. Very, very, very fat
creatures will inevitably break cutscenes a little but it's better for it
to be cosmetic stuff (creatures clipping or looking funny) and not
functional stuff (creatures spawning out of bounds or not being able to
run somewhere)

### DIFF
--- a/project/src/main/ui/overworld-ui.gd
+++ b/project/src/main/ui/overworld-ui.gd
@@ -175,12 +175,15 @@ func _on_ChatUi_pop_out_completed() -> void:
 		_update_visible()
 		emit_signal("chat_ended")
 		
-		if Level.launched_level_id:
-			Level.push_level_trail()
-			
+		if Breadcrumb.trail.size() >= 2 and Breadcrumb.trail[1] == Global.SCENE_CUTSCENE_DEMO:
+			# don't launch the level; go back to CutsceneDemo after playing the cutscene
+			Breadcrumb.pop_trail()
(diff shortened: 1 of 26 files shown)


project/src/main/ui/overworld-ui.gd
```diff
@@ -175,12 +175,14 @@ func _on_ChatUi_pop_out_completed() -> void:
 		_update_visible()
 		emit_signal("chat_ended")
 		
-		if Level.launched_level_id:
-			Level.push_level_trail()
-			
+		if Breadcrumb.trail.size() >= 2 and Breadcrumb.trail[1] == Global.SCENE_CUTSCENE_DEMO:
+			Breadcrumb.pop_trail()
```
Poobslag, line 179: Document this surprising logic

Poobslag: Added comment; 'go back to CutsceneDemo'

PR #793, opened 2021-01-29
Rewrote chat logic to key off of creature_id

Poobslag:
Chat logic used to key off of creature_name, and query the creatures in
the current scene. This caused problems:

1. CutsceneDemo did not have an actual environment, but loaded chat trees
to find metadata. This triggered the caching of 'there is no creature
named John' causing warnings and blank chat themes.

2. Multiple creatures might have the same name, particularly if the player
chooses a common name. The old code avoided this by caching some creatures
by name (Richie, Boatricia) and other creatures by id (#player#, #sensei#)

The new code keys off of creature_id, and queries the cached creature
definitions in CreatureLibrary. This is available in scenes without an
environment, avoiding the aformentioned bugs.

Assigned creature IDs for creature def json files. Now that we're looking up
creatures by their IDs, these ID fields need to be set or they're not retrieved
properly.

CreatureEditor assigns creature_id field.

Extracted StringTransformer from OldSave for reuse in other scripts.

Moved PLAYER_ID, SENSEI_ID into CreatureLibrary.

### DIFF
--- a/project/src/main/creature-library.gd
+++ b/project/src/main/creature-library.gd
@@ -5,6 +5,10 @@ Manages data for all creatures in the world, including the player.
 This includes their appearance, name and weight.
 """
 
+# unique creature ids
+const SENSEI_ID := "#sensei#"
+const PLAYER_ID := "#player#"
+
 # How many randomly generated filler creatures have their fatness tracked
 const GENERIC_FATNESS_COUNT := 150
 
@@ -16,8 +20,9 @@ const MAX_FILLER_FATNESS := 2.5
 # If set, all creatures will have a specific fatness. Used for testing cutscenes.
 var forced_fatness := 0.0
 
-var player_def: CreatureDef
-var sensei_def: CreatureDef
+# Virtual properties; values are only exposed through getters/setters
+var player_def: CreatureDef setget set_player_def, get_player_def
+var sensei_def: CreatureDef setget set_sensei_def,get_sensei_def
 
 # fatnesses by creature id
 var _fatnesses: Dictionary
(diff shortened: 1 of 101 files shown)


project/src/main/creature-library.gd
```diff
@@ -46,17 +61,73 @@ func next_filler_id() -> String:
...
+Returns a list of file paths in the specified directory.
+
+This only returns the directory's immediate children, and does not perform a tree traversal.
+
+Parameters:
+	'dir_path': The path of the directory to scan
+	
+	'file_suffix': (Optional) A suffix to append to each returned file path.
+"""
+func _file_paths(dir_path: String, file_suffix: String = "") -> Array:
```
Poobslag, line 118: Move this helper function down below the public functions.

project/src/main/ui/chat/chat-frame.gd
```diff
@@ -71,7 +71,8 @@ func play_chat_event(chat_event: ChatEvent, nametag_right: bool, squished: bool)
 	
 	# set the text and calculate how big of a frame we need
 	var chat_line_size: int = $ChatLineLabel.show_message(chat_event.text, 0.5)
-	$ChatLinePanel/NametagPanel.set_nametag_text(chat_event.who)
+	var chatter_name := PlayerData.creature_library.get_creature_def(chat_event.who).creature_name
+	$ChatLinePanel/NametagPanel.set_nametag_text(chatter_name)
```
Poobslag, line 75: `chatter_name` is an unnecessarily vague variable name; this is a creature name

project/src/main/creature-library.gd
Poobslag: Reordered private functions below public functions.

project/src/main/ui/chat/chat-frame.gd
Poobslag: Changed local variable to `creature_def` instead of `chatter_name`.

PR #803, opened 2021-02-01
Cutscene camera improvements.

Poobslag:
Camera zooms out further if characters are further apart. The old camera
behavior zoomed in too much which would clip characters off the top and
bottom of the screen.

Changed 'get_center_of_chatters' to 'get_chatter_bounding_box'. The
OverworldCamera needs to zoom out more if characters are further apart,
so it needs a rectangle and not just a coordinate.

Removed Shirts from 'hello everyone' scene.

### DIFF
--- a/project/src/main/ui/overworld-ui.gd
+++ b/project/src/main/ui/overworld-ui.gd
@@ -85,8 +85,9 @@ func is_show_version() -> bool:
 Turn the the active chat participants towards each other, and make them face the camera.
 """
 func make_chatters_face_eachother() -> void:
-	var center_of_non_player_chatters := get_center_of_chatters(
+	var chatter_bounding_box := get_chatter_bounding_box(
 			[], [ChattableManager.player, ChattableManager.sensei])
+	var center_of_non_player_chatters := chatter_bounding_box.position + chatter_bounding_box.size * 0.5
 	
 	for chatter in chatters:
 		if chatter == ChattableManager.player:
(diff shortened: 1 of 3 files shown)


project/src/main/ui/overworld-ui.gd
```diff
@@ -143,7 +144,7 @@ Quick one-line chats don't interrupt the player or zoom the camera in; the playe
 running. That's why we call them 'drive by chats'.
 """
 func is_drive_by_chat() -> bool:
-	return _current_chat_tree.events.size() == 1
+	return _current_chat_tree.events.size() <= 1 and _current_chat_tree.events.get("", []).size() <= 1
```
Poobslag, line 147: Document this change in the commit comment.

project/src/main/ui/overworld-ui.gd
```diff
@@ -119,7 +120,7 @@ Returns:
 	The center of the smallest rectangle including all characters in the current conversation. If no characters meet
 	this criteria, this returns an zero-size rectangle at (0, 0).
```
Poobslag, line 121: This method comment is no longer accurate.

project/src/main/ui/overworld-ui.gd
Poobslag: Added commit comment, `Fixed bug where the drive-by chat logic was...`

project/src/main/ui/overworld-ui.gd
Poobslag: Fixed method comment, `(Returns) the smallest rectangle including all characters...`

PR #805, opened 2021-02-02
Fixed bugs related to overworld input, buttons, focus

### LINKED ISSUE #799: It's easy to accidentally pick a dialog option when mashing through dialog
When mashing through dialog, if a dialog prompt comes up you'll just pick the first option. Maybe we should block input for a fraction of a second to make this less likely. (Or if we're already blocking input, maybe we should block it for longer.)

### LINKED ISSUE #800: Shouldn't be able to access level select from a cutscene
When a cutscene is playing, you can access the cell phone menu and select the level select screen again. This results in some strange behavior where multiple instances of the "overworld" scene are in the breadcrumb stack, and quitting backs you out to a previous overworld scene.

When a cutscene is playing, the cell phone menu should be inaccessible. You should only be able to play the level or quit.

### LINKED ISSUE #801: Quitting while cutscene is playing results in weird behavior
Playing a level with a cutscene, and quitting in the middle of the cutscene, results in some strange behavior.

The cutscene continues playing from the overworld with the characters in the wrong place, and throws a lot of warnings.

Poobslag:
Added 0.8s delay before chat choices can be picked. This avoids an issue where
the player could accidentally choose something while mashing through dialog.

Closes #799.

Restored signals for OverworldButtons, OverworldTouchButtons. Disconnecting
these signals kept the UI buttons visible during conversations and cutscenes,
allowing the player to quit or select levels while cutscenes were playing.

Closes #800. Closes #801.

Disable player input and halt their movement when the UI has focus. This
fixes some weird behavior. One example is if you were moving and a
conversation popped up, you'd keep running in that direction. Another
example is when the cell phone menu was popped up, you could navigate the
menu with the arrow keys, but if you navigated to a corner button you
could also move your character towards that corner. Conditionally
disabling player input required adding a few new signals related to the
Overworld UI.

### DIFF
--- a/project/src/main/ui/chat/chat-choices.gd
+++ b/project/src/main/ui/chat/chat-choices.gd
@@ -72,6 +72,9 @@ func show_choices(choices: Array, moods: Array, new_columns: int = 0) -> void:
 	_refresh_child_buttons()
 	
 	if choices:
+		# briefly disable input to prevent the player from mashing through chat choices accidentally
+		$EnableInputTimer.start()
+		
 		# wait for old chat choices to be deleted before grabbing focus
 		yield(get_tree(), "idle_frame")
 		for button in get_tree().get_nodes_in_group("chat_choices"):
@@ -166,6 +169,10 @@ Makes all the chat choice buttons disappear and emits a signal with the player's
 The chat choice buttons remain as children of this node so they can be animated away.
 """
 func _on_ChatChoiceButton_pressed() -> void:
+	if not $EnableInputTimer.is_stopped():
+		# don't let the player mash through chat choices accidentally
+		return
+	
 	var old_buttons := get_tree().get_nodes_in_group("chat_choices")
 
 	# determine the currently selected choice
--- a/project/src/main/ui/chat/chat-ui.gd
+++ b/project/src/main/ui/chat/chat-ui.gd
@@ -8,6 +8,7 @@ signal chat_event_played(chat_event)
 
 # emitted when we present the player with a dialog choice
 signal showed_choices
+signal chat_choice_chosen(choice_index)
 signal pop_out_completed
 
 # how long the player needs to hold the button to skip all dialog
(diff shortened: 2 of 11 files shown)


project/src/main/ui/chat/chat-choices.gd
```diff
@@ -72,6 +72,8 @@ func show_choices(choices: Array, moods: Array, new_columns: int = 0) -> void:
 	_refresh_child_buttons()
 	
 	if choices:
+		$EnableInputTimer.start()
```
Poobslag, line 75: Maybe worth documenting that this timer actually disables input for a brief period of time

project/src/main/ui/chat/chat-ui.gd
```diff
@@ -8,6 +8,7 @@ signal chat_event_played(chat_event)
 
 # emitted when we present the player with a dialog choice
 signal showed_choices
+signal chat_choice_chosen(choice_index)
```
Poobslag, line 11: Is this signal ever used?

project/src/main/world/OverworldTouchButtons.tscn
```diff
@@ -7,6 +7,7 @@
 modulate = Color( 1, 1, 1, 0.627451 )
 anchor_right = 1.0
 anchor_bottom = 1.0
+mouse_filter = 2
```
Poobslag, line 10: This fixes a bug where chat buttons were unclickable; you should mention that in the commit comment.

project/src/main/ui/chat/chat-ui.gd
Poobslag: Yes; this signal is used by OverworldUi.

project/src/main/world/OverworldTouchButtons.tscn
Poobslag: Expanded commit comment; `Fixed bug where chat choices could not be clicked on overworld...`

project/src/main/ui/chat/chat-choices.gd
Poobslag: Added code comment, `briefly disable input to prevent the player...`

PR #809, opened 2021-02-08
Scenarios support chefs, customers. Added customer metabolism

Poobslag:
Added customer_ids property to worlds.json, so the level can have a
different character and customer. Not all levels will involve serving the person
you talk to; sometimes you'll talk to employees, or someone will want you to
cater for their friends.

Added weight_gain_scale, metabolism_scale settings. Rhonk gains weight very
slowly and sheds weight very quickly. This is for thematic reasons.

Changed combo_break rule logic so that a value of 999,999 disables combo breaks
entirely. Previously, it was technically possible for someone to break
their combo by dropping 1,000,000 pieces, if they were willing to play the
same level for a week.

combo_break 999,999 rule now includes a special case, just in case the
player actually drops 1,000,000 pieces (!?)

Scenarios can have different chefs, as specified by the optional 'chef_id'
parameter

ChatLibrary.chat_tree_for_creature_id no longer pushes a warning for
non-existent chat trees. Some levels don't have cutscenes currently, and
that's not error or warning-worthy. This can be re-enabled with a flag
later once cutscenes are added.

### DIFF
--- a/project/src/main/ui/chat/chat-library.gd
+++ b/project/src/main/ui/chat/chat-library.gd
@@ -6,6 +6,9 @@ Dialog is stored as a set of json resources. This class parses those json resour
 into the UI.
 """
 
+# Missing chat tree warnings are currently disabled. Many levels don't have cutscenes.
+const WARN_ON_MISSING_CHAT_TREE := false
+
 """
 Loads a conversation for the specified creature.
 
@@ -32,6 +35,8 @@ func load_chat_events_for_creature(creature: Creature, forced_level_id: String =
 
 """
 Returns the chat tree for the specified creature.
+
+Returns null if the chat tree cannot be found.
 """
 func chat_tree_for_creature_id(creature_id: String, forced_level_id: String = "") -> ChatTree:
 	var creature_def: CreatureDef = PlayerData.creature_library.get_creature_def(creature_id)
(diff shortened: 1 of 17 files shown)


project/src/main/ui/chat/chat-library.gd
```diff
@@ -62,9 +69,11 @@ func chat_tree_for_creature_def(creature_def: CreatureDef, state: Dictionary) ->
 	elif FileUtils.file_exists(level_dialog_path):
 		chat_tree = load_chat_events_from_file(level_dialog_path)
 	else:
-		push_warning("Failed to load chat tree '%s' for creature '%s'.\nCould not find file '%s' or '%s'" % \
+		if WARN_ON_MISSING_CHAT_TREE:
+			push_warning("Failed to load chat tree '%s' for creature '%s'.\nCould not find file '%s' or '%s'" % \
 				[chosen_dialog, creature_id, creature_dialog_path, level_dialog_path])
 	
+
```
Poobslag, line 76: Remove extra line break.

project/src/main/world/creature/creature-def.gd
```diff
@@ -68,6 +68,9 @@ var chat_selectors: Array
 # how fat the creature's body is; 5.0 = 5x normal size
 var min_fatness := 1.0
 
+var weight_gain_scale := 1.0
+var metabolism_scale := 1.0
```
Poobslag, line 72: Add field comments.

project/src/main/ui/chat/chat-library.gd
Poobslag: Removed

project/src/main/world/creature/creature-def.gd
Poobslag: Added comments, `how fast the creature should gain weight...`

PR #811, opened 2021-02-11
Minor enhancements, bug fixes and refactorings for puzzles and cutsce…

Poobslag:
…nes.

== Enhancements

The 'Start/Back' buttons in a puzzle are now more complex. Upon clearing a
puzzle, these change to 'Retry/Continue' and the continue button is focused to
encourage the player to move on. Practice puzzles are treated differently,
and the retry button is focused instead.

Aborting a puzzle without dropping pieces is no longer recorded as a loss, but
as a 'NONE'. There is no losing sound effect and no delay for the buttons
reappearing.

CutsceneDemo has toggle for long names.

== Bug fixes

Fixed bug where levels were always pausing for 4.2 seconds after
finishing, instead of alternating between 4.2/2.2 seconds depending on the
outcome of the puzzle.

== Refactorings

Moved LOCATION_SCENE_PATHS_BY_ID from level.gd into ChatTree.

Removed PuzzleScore.LOST, PuzzleScore.WON, PuzzleScore.FINISHED
convenience constants. The new PuzzleScore.NONE constant would have been
ambiguous as a convenience constant, so I just removed them all.

Renamed some ChatLibrary methods. ChatLibrary was inconsistent about
refering to things as 'load_' or 'chat_tree' or 'chat_events' but it's now
more consistent.

### DIFF
--- a/project/src/main/puzzle/level/level.gd
+++ b/project/src/main/puzzle/level/level.gd
@@ -11,15 +11,22 @@ This class needs to be a node because it provides utility methods which utilize 
 # emitted after the level has customized the puzzle's settings.
 signal settings_changed
 
+# emitted when the 'level_state' field changes, such as when starting or clearing a level.
+signal level_state_changed
+
+enum LevelState {
+	NONE, # the level hasn't been launched
+	BEFORE, # the level has been launched, but the hasn't yet been finished successfully
+	AFTER # the level has been finished successfully. The player didn't lose or give up
+}
+
 # The mandatory tutorial the player must complete before playing the game
 const BEGINNER_TUTORIAL := "tutorial/basics_0"
 const TUTORIAL_WORLD_ID := "tutorial"
 
-# Scene paths corresponding to different ChatTree.location_id values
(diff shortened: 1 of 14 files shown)


project/src/main/puzzle/level/level.gd
```diff
@@ -11,15 +11,22 @@ This class needs to be a node because it provides utility methods which utilize
 # emitted after the level has customized the puzzle's settings.
 signal settings_changed
 
+# emitted when the 'level_state' field changes, such as when starting or clearing a level.
+signal level_state_changed
+
+enum LevelState {
+	NONE,
+	BEFORE, # the level hasn't been started
```
Poobslag, line 19: Maybe clarify this a bit; `NONE` = the level hasn't been launched yet; `BEFORE` = level has been launched but hasn't been played yet

Poobslag: `AFTER` = the level has been finished successfully

Poobslag: Updated enum constants to be more descriptive; `the level hasn't been launched`

PR #837, opened 2021-03-06
Added language menu option. Expanded message files.

Poobslag:
Added new language settings to settings menu. The player's chosen
language is saved/loaded with their other settings.

Message files include song titles, gameplay hints, and other overlooked phrases,
so that they can be translated.

### DIFF
--- a/project/src/main/ui/settings-language.gd
+++ b/project/src/main/ui/settings-language.gd
@@ -1,0 +1,19 @@
+extends HBoxContainer
+"""
+UI control for changing the game's language.
+"""
+
+onready var _option_button: OptionButton = $OptionButton
+
+func _ready() -> void:
+	# populate the dropdown with the names of available locales
+	for locale in TranslationServer.get_loaded_locales():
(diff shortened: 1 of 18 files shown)


project/src/main/ui/settings-language.gd
```diff
@@ -0,0 +1,12 @@
+extends HBoxContainer
```
Poobslag, line 1: Add a class-level comment.

project/src/main/ui/settings-tab-container.gd
```diff
@@ -0,0 +1,14 @@
+extends TabContainer
```
Poobslag, line 1: Add a class-level comment.

project/src/main/ui/settings-language.gd
Poobslag: Added comment, `UI control for changing the game's language.`

project/src/main/ui/settings-tab-container.gd
Poobslag: Added comment, `UI control for changing the game's language.`.

PR #841, opened 2021-03-14
Improved CreatureVisuals performance by removing viewports.

Poobslag:
CreatureVisuals had three viewports with custom shaders for rendering
the creature's torso. Instead of rendering these in separate viewports, I've
created a custom node which does some polygon math so that the torso can
be drawn directly to the Creature's main viewport.

Benchmarks suggest this improves performance from 2x-5x. Instead of
stress test performance degrading from 700 fps down to 25 fps, it now
starts at 1200 fps and stays as high as 170 fps even with 45 creatures on the
screen.

Set main creature viewport's mode to '2D'. This bumps performance up
another 25% or so.

Check for null material when applying shader keys.

Removed some problematic keys from Terpion. Terpion had some keys
setting specific shader parameters. No other creatures have this, and
the material no longer exists potentially causing null pointer
exceptions.

Made some renames to avoid confusion with the new Body scene. The code
used to alternate between the terms 'Belly' and 'BodyColors' and with
the new parent node for rendering these nodes in a special way, it was
necessary to come up with a name for it. So instead of:

CreatureVisuals/Body
CreatureVisuals/Body/NeckBlend
CreatureVisuals/BodyColors
CreatureVisuals/BodyShadows

The new node layout is like this:

CreatureVisuals/Body
CreatureVisuals/Body/BodyShape
CreatureVisuals/Body/Belly
CreatureVisuals/Body/Shadows
CreatureVisuals/Body/NeckBlend

Neckblend is now drawn as a path instead of a sprite.


project/src/main/world/creature/creature-curve.gd
```diff
@@ -21,10 +24,18 @@ export (bool) var _save_curve: bool setget save_curve
 export (Array) var curve_defs: Array setget set_curve_defs
 
 # if false, the curve is made invisible when the creature faces away from the camera.
-export (bool) var visible_when_facing_north: bool = true
+export (bool) var drawn_when_facing_north: bool = true
+
+# if false, the curve is made invisible when the creature faces toward the camera.
```
Poobslag, line 29: Maybe these comments shouldn't say "visible". Something more like, "The body will not render this curve when the creature faces toward the camera"

project/src/main/world/restaurant/RestaurantView.tscn
```diff
@@ -215,6 +215,6 @@ margin_bottom = 367.0
 margin_left = -60.0
 margin_right = 60.0
 custom_styles/panel = SubResource( 4 )
-[connection signal="current_creature_index_changed" from="RestaurantViewport/Scene" to="CustomerCameraMover" method="_on_RestaurantScene_current_creature_index_changed"]
 [connection signal="current_creature_index_changed" from="RestaurantViewport/Scene" to="." method="_on_RestaurantScene_current_creature_index_changed"]
+[connection signal="current_creature_index_changed" from="RestaurantViewport/Scene" to="CustomerCameraMover" method="_on_RestaurantScene_current_creature_index_changed"]
```
Poobslag, line 219: Revert this.

Poobslag: Reverted.

project/src/main/world/creature/creature-curve.gd
Poobslag: Changed to `if false, the body will not render this curve when the creature faces away from the camera.`

PR #845, opened 2021-03-20
Removed overworld goop/ground viewports.

### LINKED ISSUE #742: Border/jitter on overworld goop texture when moving, zooming out
image

When the overworld camera moves or zooms out, there's a subtle margin. It's visible in this screenshot on the left side.

Additionally in cutscenes which cause the camera to zoom out when a character leaves (such as `marsh/hello_bones_100`) there's a weird rubberband effect where the textures bounce back and forth.

I believe this is because the "ScrewportTexRect" scaling and position lags behind everything else by a frame. One easy fix might be to give it a 20 pixel margin.

### LINKED ISSUE #826: Avoid viewport when rendering overworld
Based on the performance investigation documented in #767, the overworld's goop viewport is a performance bottleneck. This causes FPS to degrade from ~300 FPS to ~225 FPS on my desktop PC. Issue #742 describes some visual artifacts resulting from these viewports as well.

We should find a way to render the overworld without these viewports. We could use a conventional sprite sheet for these overworld tiles.

Poobslag:
The overworld used to have two viewports for rendering the goop and
ground. It would render the goop to one ViewportTexture, then use
combine that texture with a few traditional image-based textures to
render the ground.

Instead, it now applies all of this rendering logic in a shader on the
tileset. It does not require any viewports.

Surprisingly, this does not have a significant impact on performance.
There is a performance benefit in avoiding the two viewports, but a
performance cost in applying a shader to 500+ tiles.

Closes #826. Closes #742.

Increased the cell_quadrant_size for the overworld. The goop shader has
visible seams at the quadrant edges, and I couldn't find a way to avoid
this. The easiest fix is simply to avoid having goop tiles overlay
seams.

Removed OverworldObstacles, OverworldDemo. These were only kept separate
because of the difficulty of working with the viewport-based rendering
approach.

Resolved a possible NPE when rendering a body before its shadow node
exists.


project/src/main/world/goop-ground.shader
```diff
@@ -0,0 +1,85 @@
...
+ * component with a flowing goop pattern. Black pixels are also replaced with a solid color.
+ *
+ * This approach allow a sprite to be recolored with up to four channels, and the recoloring can be handled by the
+ * graphics card. It's performant and looks reasonable even if the sprite is resized before being passed off to the
+ * graphics card.
+ */
+shader_type canvas_item;
+
+// squash factor to apply an isometric effect, or to maintain the goop's aspect ratio
+const vec2 squash_factor = vec2(0.5, 1.0);
```
Poobslag, line 12: Constants use `UPPER_CASE_SNAKE_CASE`.

project/src/main/world/goop-ground.shader
```diff
@@ -0,0 +1,85 @@
...
+float noise(vec2 p) {
+	return textureLod(noise_texture, p, 2.0).r;
+}
+
+vec4 goop_color() {
+    vec2 uv0 = local * squash_factor / vec2(textureSize(noise_texture, 0));
+	vec2 uv1 = uv0 + vec2(38.913, 81.975) + time * SCROLL_SPEED;
+	vec2 uv2 = uv0 * vec2(4.0, 4.0) + noise(1.2 * uv1) + vec2(time * DISTORTION_SPEED, 0.0);
+
+    // calculate a number evenly distributed in the range [0.0, 1.0) based on noise
```
Poobslag, line 59: fix the formatting of this method; indentation isn't consistent

project/src/main/world/goop-ground.shader
```diff
@@ -0,0 +1,85 @@
...
+	
+	vec4 red = texture(red_texture, local / vec2(textureSize(red_texture, 0)));
+	vec4 blue = texture(blue_texture, local / vec2(textureSize(blue_texture, 0)));
+	
+	// mix black/red/blue/green colors sequentially. if all rgb components are equal, we first mix in 100% of the red
+	// color, then 50% of the green color, then 33% of the blue color.
+	float black_amount = max(0.0, 1.0 - rgba_in.r - rgba_in.g - rgba_in.b);
+	vec4 rgb_out = black;
+	rgb_out = mix(rgb_out, red, rgba_in.r / max(0.00001, rgba_in.r + black_amount));
+	rgb_out = mix(rgb_out, goop_color(), rgba_in.g / max(0.00001, rgba_in.g + rgba_in.r + black_amount));
```
Poobslag, line 79: move this above, `vec4 green = goop_color()` for consistency

project/src/main/world/goop-ground.shader
Poobslag: Changed to `SQUASH_FACTOR`

project/src/main/world/goop-ground.shader
Poobslag: Fixed; code was alternating between tabs and spaces before

project/src/main/world/goop-ground.shader
Poobslag: Moved `green` variable alongside the others for consistency

PR #846, opened 2021-03-23
Fixed localization bugs: tutorial, fuzzy match

Poobslag:
Fixed 'mega_pause_charaters' typo. This method was renamed to
mega_lull_characters but the reference was not changed. The result was
that triggering the tutorial easter egg would make the tutorial
unfinishable.

Fixed bug where unsupported locale caused menu to be inconsistent.
Specifically, when testing the game my locale was 'en_US' but this did
not match our supported locale of 'en', so it set the selected language
menu item to 'Spanish' even though the game was displayed in english.

The game now matches 'en' to the locale 'en_US', and 'es' to the local
'es_MX' in the settings menu. It does this with an algorithm similar to
the algorithm used by Godot for locale matching.

### DIFF
--- a/project/src/main/ui/settings-language.gd
+++ b/project/src/main/ui/settings-language.gd
@@ -3,6 +3,8 @@ extends HBoxContainer
 UI control for changing the game's language.
 """
 
+const DEFAULT_LOCALE := "en"
+
 onready var _option_button: OptionButton = $OptionButton
 
 func _ready() -> void:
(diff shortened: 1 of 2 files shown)


project/src/main/ui/settings-language.gd
```diff
@@ -11,7 +13,40 @@ func _ready() -> void:
 		_option_button.add_item(TranslationServer.get_locale_name(locale))
 	
 	# update the option button's selection to the current locale
-	_option_button.selected = TranslationServer.get_loaded_locales().find(TranslationServer.get_locale())
+	_option_button.selected = TranslationServer.get_loaded_locales().find(_current_loaded_locale())
```
Poobslag, line 16: It might be worth spitting out an error if it doesn't exist. We don't want this to bite us in the future.

Poobslag: Added error, `Locale '%s' was not in the list of loaded locales`

PR #855, opened 2021-03-27
Fix ghost piece toggle bug, splash fade bug.

### LINKED ISSUE #853: Toggling ghost piece mid-puzzle results in strange behavior
Toggling the ghost-piece setting on and off while playing a puzzle results in some strange behavior: The ghost piece will gradually fall beneath the bottom of the well.

Poobslag:
Toggling the ghost piece in the middle of a puzzle caused some unusual
behavior. The shadow offset would retain the value for the most recent
shadow piece. So the ghost piece stuck around, it was just in a silly
place. We now enable/disable the ghost piece correctly and respond to
the player changing their settings mid-puzzle.

Closes #853.

Fixed a bug where clicking 'Play' while the splash screen was still
fading in resulted in all the subsequent menus fading in and out as
well.

### DIFF
--- a/project/src/main/gameplay-settings.gd
+++ b/project/src/main/gameplay-settings.gd
@@ -3,8 +3,17 @@ class_name GameplaySettings
 Manages settings which control the gameplay.
 """
 
+signal ghost_piece_changed(value)
+
 # 'true' if a ghost piece should be shown during the puzzle sections.
-var ghost_piece := true
+var ghost_piece := true setget set_ghost_piece
+
+func set_ghost_piece(new_ghost_piece: bool) -> void:
+	if ghost_piece == new_ghost_piece:
+		return
+	ghost_piece = new_ghost_piece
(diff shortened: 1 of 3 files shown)


project/src/main/gameplay-settings.gd
```diff
@@ -3,8 +3,17 @@ class_name GameplaySettings
 Manages settings which control the gameplay.
 """
 
+signal ghost_piece_changed(value)
+
 # 'true' if a ghost piece should be shown during the puzzle sections.
-var ghost_piece := true
+var ghost_piece := true setget set_ghost_piece
+
+func set_ghost_piece(value: bool) -> void:
```
Poobslag, line 11: This setter doesn't adhere to our conventions; `set_ghost_piece(new_ghost_piece)`

Poobslag: Fixed; `func set_ghost_piece(new_ghost_piece: bool) -> void:`

PR #860, opened 2021-03-29
Added levelup particles. Combos use Particles2D.

### LINKED ISSUE #819: Add a visual cue for speedups
There's an audio cue when the game speeds up. Add an accompanying visual cue as well.

This visual cue would prevent people playing without sound from being startled by speedups.

Maybe make the 'milestone bar' more visible, or make it ripple when it hits a milestone.

Poobslag:
Added particle effect when levelling up. This provides visual feedback for
levelups for players playing without sound.

Closes #819.

Converted combo counter from CPUParticles2D to Particles2D. Particles2D
should be more performant and there's no reason to use the CPU for this.

### DIFF
--- a/project/src/main/puzzle/milestone-hud.gd
+++ b/project/src/main/puzzle/milestone-hud.gd
@@ -21,9 +21,11 @@ const LEVEL_COLOR_5 := Color("b948b9")
 onready var _progress_bar: ProgressBar = $ProgressBar
 onready var _desc: Label = $Desc
 onready var _value: FontFitLabel = $Value
+onready var _progress_bar_particles: Control = $ProgressBarParticles
 
 func _ready() -> void:
 	PuzzleScore.connect("game_prepared", self, "_on_PuzzleScore_game_prepared")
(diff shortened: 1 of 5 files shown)


project/src/main/puzzle/milestone-hud.gd
```diff
@@ -120,5 +122,26 @@ func _on_PuzzleScore_game_prepared() -> void:
...
+We emit particles when the player levels up.
+
+This provides visual feedback for people playing without sound.
+"""
+func _on_PuzzleScore_speed_index_changed(value: int) -> void:
+	if value == 0:
+		# initializing the level; don't emit any particles
+		return
+	
+	var particles_material: ParticlesMaterial = _progress_bar_particles.get_child(0).process_material
```
Poobslag, line 135: It's a little strange that we only modify one of the children; worth documenting

Poobslag: Added comment, `All of the Particles2D share the same GradientTexture so we only need to modify one.`

PR #862, opened 2021-03-29
Restored broken creature_visuals listeners.

### LINKED ISSUE #861: The hitboxes of the big monsters are small
image

The hitboxes of the big monsters are small.

They used to scale with the creature's size, so something got broken at some point.

Poobslag:
ChatIconHook and CollisionShape2D require a CreatureVisuals reference to
position and size themselves. This reference was broken during the
recent 'low quality creature visuals' change, so the chat icons would
always be in the same place, and fat creatures would still have small
hit boxes.

Closes #861.

### DIFF
--- a/project/src/main/world/creature/creature.gd
+++ b/project/src/main/world/creature/creature.gd
@@ -105,6 +105,7 @@ var _creature_outline: CreatureOutline
 
 onready var _creature_sfx: CreatureSfx = $CreatureSfx
 onready var _collision_shape: CollisionShape2D = $CollisionShape2D
+onready var _chat_icon_hook: RemoteTransform2D = $ChatIconHook
 onready var _fade_tween: Tween = $FadeTween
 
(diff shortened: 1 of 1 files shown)


project/src/main/world/creature/creature.gd
```diff
@@ -119,6 +119,8 @@ func _ready() -> void:
 	add_child(_creature_outline)
 	
 	creature_visuals = $CreatureOutline.creature_visuals
+	$ChatIconHook.creature_visuals_path = $ChatIconHook.get_path_to(creature_visuals)
+	$CollisionShape2D.creature_visuals_path = $CollisionShape2D.get_path_to(creature_visuals)
```
Poobslag, line 123: Use the `_collision_shape` field, and consider creating a new `chat_icon_hook` field.

Poobslag: Rewrote logic to use `_collision_shape`, `_chat_icon_hook` fields.

PR #878, opened 2021-04-11
Keybinds, 'Save + Quit' messages support localization

Poobslag:
Replaced 'quit_message' with 'quit_type' so that the resulting strings
can be localized.

Because Godot #4140 (https://github.com/godotengine/godot/issues/4140) is not
yet implemented there is no way to get a localized version of
OS.get_scancode_string. I've updated localization-demo.gd to extract all OS
scancodes so we can translate them ourselves.

preset-keybind-row.gd now uses the label 'keybind_value' instead of
'value' because this is less ambiguous for pybabel-extract.sh. I'm
nervous that there could be all sorts of unrelated 'values' which would
be extracted otherwise (spinner values, slider values, etc)


project/src/main/ui/menu/settings-menu.gd
```diff
@@ -10,12 +10,16 @@ signal show
...
-const GIVE_UP := "Give Up"
+enum QuitType {
+	QUIT, SAVE_AND_QUIT, GIVE_UP
+}
+
+const QUIT := QuitType.QUIT
+const SAVE_AND_QUIT := QuitType.SAVE_AND_QUIT
+const GIVE_UP := QuitType.GIVE_UP
 
 # The text on the menu's quit button
```
Poobslag, line 21: Fix this comment

PR #882, opened 2021-04-17
Bugfixes: sensei weight gain, class instance is gone, crumb count

Poobslag:
Fixed bug where sensei gains weight in tutorials.

Fixed 'class instance is gone' errors with food.

Fixed crumb count calculation. Calculation was producing an integer
division warning, and producing incorrect results because of an
unintuitive behavior of rand_range(). var r: int = rand_range(1, 2) does not
return a random number from 1 to 2, it returns 1.

### DIFF
--- a/project/src/main/puzzle/food-item.gd
+++ b/project/src/main/puzzle/food-item.gd
@@ -42,6 +42,12 @@ var velocity := Vector2(0, -25)
 # The creature who this food will fly towards
 var customer: Creature
 
+# Incremented as the customer changes to track the intended recipient of each food item.
+var customer_index: int
+
+# An enum from FoodType corresponding to the food to show
+var food_type: int setget set_food_type
+
 # Food items pulse and rotate. This field is used to calculate the pulse/rotation amount
 var _total_time := 0.0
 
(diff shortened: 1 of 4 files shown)


project/src/main/puzzle/food-item.gd
```diff
@@ -42,6 +42,11 @@ var velocity := Vector2(0, -25)
 # The creature who this food will fly towards
 var customer: Creature
 
+# Incremented as the customer changes to track the intended recipient of each food item.
+var customer_index: int
+
+var food_type: int setget set_food_type
```
Poobslag, line 48: Add a comment

project/src/main/world/creature/creature.gd
```diff
@@ -601,7 +601,12 @@ func _on_FadeTween_tween_all_completed() -> void:
 Converts a score in the range [0.0, 5000.0] to a fatness in the range [1.0, 10.0]
 """
 func score_to_fatness(in_score: float) -> float:
-	return sqrt(1 + in_score / (50.0 / weight_gain_scale))
+	var result: float
+	if creature_id == CreatureLibrary.SENSEI_ID:
+		result = base_fatness
```
Poobslag, line 606: Add a comment

project/src/main/puzzle/food-item.gd
Poobslag: Added comment: `An enum from FoodType corresponding to the food to show`

project/src/main/world/creature/creature.gd
Poobslag: Added comment: `# tutorial sensei doesn't gain weight`

PR #883, opened 2021-04-17
Replaced json dialog with chatscript.

### LINKED ISSUE #782: Create cutscene dialog tools
We have a lot of cutscenes to make, and we will have many, many, many more. We want this process to be efficient.

We should create whatever tools we need to make writing dialog easier, without burdening ourself with extra development work. This could be a standalone UI, an extension of the Godot editor, some specialized tool nodes, or command-line scripts.

For example, instead of needing to write something like this:
```
      "" : [
        {
          "who": "Bort",
          "text": "Wow!/ ./././You finally did it?/ Does this count as our first conversation!?",
          "mood": "think0",
          "links": [
            {
              "yes": "Yeah, definitely!"
            },
            {
              "maybe": "Maybe? I'm not sure!"
            },
            {
              "no": "I don't think so..."
            }
          ]
        }
      ],
      "yes" : [
        {
          "who": "#player#",
          "text": "Yeah!\nThis is definitely a real conversation./ It has branches and everything!",
          "mood": "smile0"
        },
        {
          "who": "Bort",
          "text": "Congratulations!/ Wow!",
          "mood": "smile0",
          "links": ["nice"]
        }
      ],
```

It might be preferable to have a condensed format like this:

```
b: @_@ Wow!/ ./././You finally did it?/ Does this count as our first conversation!?
[yes] Yeah, definitely!
[maybe] Maybe? I'm not sure!
[no] I don't think so...

[yes]
p: ^_^ Yeah!\nThis is definitely a real conversation./ It has branches and everything!
b: ^_^ Congratulations!/ Wow!
[nice]
```

Poobslag:
Converted all dialog and cutscenes from json to chatscript. Chatscript
is a condensed format which is less versatile but much easier to write.

Closes #782.

Removed support for creatures with inline dialog.

Pulled bort's dialog out of his creature.json. I think this may have
been a proof of concept for the 'all in one file' type creature, which
may eventually be supported by the creature editor. But right now it's a
lot of complexity for complexity's sake -- if we go the 'all in one
file' approach we'll need to think about this case then.

load_chat_events -> load_chat_tree; 'ChatEvents' was renamed to
'ChatTree' a long time ago but this one reference was not changed.

Removed unused 'no-vegetables', 'level-1', 'level-2' conversations.
These were redundant with the cutscenes and never appeared in-game.


project/src/main/ui/chat/chat-library.gd
```diff
@@ -283,48 +266,37 @@ func add_mega_lull_characters(s: String) -> String:
 	return transformer.transformed
 
 
+func _creature_dialog_chat_path(creature_id: String, chat_id: String) -> String:
```
Poobslag, line 269: Rename this back to `_creature_dialog_path`

project/src/main/ui/chat/chatscript-parser.gd
```diff
@@ -0,0 +1,378 @@
...
+			_character_aliases[character_alias] = character_name
+
+
+"""
+Parser state for parsing dialog data (chat events and branches)
+"""
+class DialogState extends AbstractState:
+	var _character_aliases: Dictionary
+	var _event: ChatEvent
+	var _branch_key := ""
```
Poobslag, line 128: Document these fields

project/src/main/ui/chat/chatscript-parser.gd
```diff
@@ -0,0 +1,378 @@
...
+		else:
+			result = DEFAULT
+		return result
+
+
+"""
+Parser state for parsing character data (participants in a conversation).
+"""
+class CharactersState extends AbstractState:
+	var _character_aliases: Dictionary
```
Poobslag, line 87: Document this field

project/src/main/ui/chat/chat-library.gd
Poobslag: Renamed to `_creature_dialog_path`

project/src/main/ui/chat/chatscript-parser.gd
Poobslag: Added comment, `Creature aliases used in chatscript dialog. DialogState references this dictionary when parsing dialog`

project/src/main/ui/chat/chatscript-parser.gd
Poobslag: Added comment, `Creature aliases used in chatscript dialog. DialogState references this dictionary when parsing dialog`

PR #906, opened 2021-05-03
Rotate cw/ccw. Elevation is a float. Frame drop ui fixes.

Poobslag:
We now use consistent terminology throughout our code: 'rotate cw', 'rotate
ccw', 'rotate 180'. '180' is occasionally represented as 'flip' in contexts
where the number 180 causes problems, such as variable names. These are also
still presented as 'Rotate Right' and 'Rotate Left' in the tutorial, because
'Rotate Counterclockwise' is far too wide for the tutorial box.

Removed unused PieceType.Orientation enum.

Creature.elevation and CreatureChatIconHook.elevation are now floats.

FrameDrops label disables process/physics_process when invisible, for
performance and simplicity. Changed a few overworld UI and puzzle UI
panels to ignore mouse input; enabling the 'BIGFPS' puzzle cheat made
some parts of the buttons unclickable because the panels were eating
mouse inputs.


project/src/main/puzzle/tutorial/skill-tally-item.gd
```diff
@@ -9,7 +9,7 @@ This is used during tutorials so the player can see what to do.
 # if 'true' the skill tally will be shown as '58%' instead of '52/90'
 export (bool) var show_as_percent: bool
 
-# description of the skill being tallied such as 'Rotate Left'
+# description of the skill being tallied such as 'Rotate Counterclockwise'
```
Poobslag, line 12: Revert this; we changed its name back

project/src/main/ui/scene-transition.gd
```diff
@@ -65,8 +65,8 @@ Parameters:
 """
 func replace_trail(path: String, skip_transition: bool = false) -> void:
 	if skip_transition \
-	        or not get_tree().get_nodes_in_group("scene_transition_covers") \
-	        or "::" in path:
+			or not get_tree().get_nodes_in_group("scene_transition_covers") \
+			or "::" in path:
```
Poobslag, line 69: This is a valuable change, but consider a delinting rule

project/src/main/puzzle/tutorial/skill-tally-item.gd
Poobslag: Reverted.

project/src/main/ui/scene-transition.gd
Poobslag: Added a delinting rule.

PR #914, opened 2021-05-08
Improve starting customer logic, food targeting efficiency

Poobslag:
The old starting customer logic had conflicting logic in the RestaurantView and
Puzzle scripts. 'Puzzle' would reload all the new creatures, and then
'RestaurantView' would try to scroll to a new creature if they had been fed, but
since all creatures had just been reloaded, it would always just scroll to a new
creature.

The new starting customer logic is more robust, and handles edge cases
such as restarting a puzzle when no creatures have been fed, restarting
a story puzzle when the camera is on the starting creature, and so on.

FoodItems target calculations are now cached, instead of being
recalculated 20 times for 20 different food items. They're all going to
the same location so it's silly to perform multiple lerps/transforms
over and over to get the same answer.

### DIFF
--- a/project/src/main/puzzle/food-item.gd
+++ b/project/src/main/puzzle/food-item.gd
@@ -54,9 +54,8 @@ var _total_time := 0.0
 # Tweens the food's position to fly into the customer's mouth.
 var _flying_tween: Tween
 
-# This object, method and arg_array correspond to a callback which return the position of the customer's mouth.
-var _target_pos_object: Object
-var _target_pos_method: String
+# This func ref and array correspond to a callback which return the position of the customer's mouth.
+var _get_target_pos: FuncRef
 var _target_pos_arg_array: Array
 
 # The position the food is flying from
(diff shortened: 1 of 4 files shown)


project/src/main/puzzle/food-item.gd
```diff
@@ -129,10 +128,9 @@ Parameters:
 	
 	'target_pos_arg_array': The parameters of the callback function which returns the position of the customer's mouth.
```
Poobslag, line 129: The parameters changed, the method docs need to update too

project/src/main/puzzle/food-items.gd
```diff
@@ -76,26 +85,30 @@ func add_food_item(cell: Vector2, food_type: int) -> void:
 
 """
 Callback function which returns the coordinate of the customer's mouth relative to the FoodItems viewport.
+
+This is an expensive calculation which is needed once per frame per food item, so we employ a cache
```
Poobslag, line 89: Add a period to the end of this sentence.

Poobslag: Added punctuation.

PR #918, opened 2021-05-09
Avoid crumb performance hiccup. Squishes trigger DAS.

Poobslag:
Initializing the first crumb cluster sometimes causes a performance hiccup. I've
introduced a pool so that this hiccup occurs on load, rather than during
gameplay.

Squish moves trigger DAS. This fixes misdrops related to rocket drops,
which occurred when trying to squish a piece into a nook while playing
quickly or during 20G.

### DIFF
--- a/project/project.godot
+++ b/project/project.godot
@@ -204,6 +204,11 @@ _global_script_classes=[ {
 "language": "GDScript",
 "path": "res://src/main/world/creature/creature-visuals.gd"
 }, {
+"base": "Node2D",
+"class": "CrumbCluster",
+"language": "GDScript",
+"path": "res://src/main/puzzle/crumb-cluster.gd"
+}, {
 "base": "Reference",
 "class": "DnaAlternatives",
 "language": "GDScript",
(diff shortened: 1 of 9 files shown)


project/src/main/world/creature/CreatureVisuals.tscn
```diff
@@ -675,11 +675,11 @@ script = ExtResource( 45 )
 creature_visuals_path = NodePath("../../..")
```
Poobslag, line 675: Revert this file.

project/src/main/puzzle/CrumbCluster.tscn
```diff
@@ -77,4 +77,5 @@ process_material = SubResource( 6 )
 texture = ExtResource( 4 )
 
 [node name="Timer" type="Timer" parent="."]
+
```
Poobslag, line 80: Revert this.

project/src/main/world/creature/CreatureVisuals.tscn
Poobslag: Reverted.

project/src/main/puzzle/CrumbCluster.tscn
Poobslag: Reverted.

PR #923, opened 2021-05-23
Added cake block obstacles on the overworld.

Poobslag:
Overworld corner covers now cover the upper corner instead of a lower
corner. This avoids a layering issue when the corner covers are y
sorted. Before, the corner covers were being y sorted above the tiles
they were meant to cover.


project/src/main/world/overworld-corner-map.gd
```diff
@@ -14,15 +21,16 @@ func _process(_delta: float) -> void:
...
 		for cell in _parent_map.get_used_cells():
-			if _parent_map.get_cellv(cell) != 0:
+			var cell_index := _parent_map.get_cellv(cell)
+			if not cell_index in cornerable_tiles:
 				# not a merrymellow marsh cell; don't fill in the corner
 				continue
 			
-			var pab := _pab(cell)
+			var pab := _pab(cell_index, cell)
 			# check for corner connected down and right
```
Poobslag, line 30: This comment is no longer accurate.

Poobslag: Updated; `check for corner connected up and left`

PR #932, opened 2021-06-02
Changed indoor floor tiles from rectangles to parallelograms.

Poobslag:
Having the interior tiles and collision shapes as rectangles resulted in a
problem when trying to design a vertical countertop. This would either need to
be diagonal and not align with the tilemap, or it would need to be vertical and
drawn strangely.

This affects the indoor floor tiles, shadows, tilemap and collision
shapes. Assets such as tables, chairs and walls were not redrawn. Table
and chair collision shapes remain as rectangles, as this feels more
intuitive for now.


project/src/main/world/OverworldIndoors.tscn
```diff
@@ -146,10 +149,13 @@ __meta__ = {
 
 [node name="ObstacleMap" type="TileMap" parent="World/Obstacles"]
 scale = Vector2( 0.5, 0.5 )
+mode = 2
 tile_set = ExtResource( 5 )
 cell_size = Vector2( 128, 64 )
+cell_custom_transform = Transform2D( 128, 0, -96, 128, 0, 0 )
 cell_tile_origin = 1
 cell_y_sort = true
+show_collision = true
```
Poobslag, line 158: Disable the `show_collision` flag.

project/src/main/world/OverworldIndoors.tscn
```diff
@@ -690,7 +696,6 @@ dna = {
 "shader:TailZ1:red": Color( 0.560784, 0.105882, 0.129412, 1 ),
 "tail": "5"
 }
-elevation = 150
```
Poobslag, line 693: Why was this flag removed? Is the creature still drawn correctly?

project/src/main/world/OverworldIndoors.tscn
Poobslag: Disabled.

project/src/main/world/OverworldIndoors.tscn
Poobslag: Restored `elevation-150` value; the creature was being drawn on the floor instead of on the stool.

PR #935, opened 2021-06-13
Added grill graphics, grill autotiling logic.

Poobslag:
Rearranged kitchen so cutscenes have space.


project/src/demo/world/CutsceneDemo.tscn
```diff
@@ -143,6 +143,7 @@ window_title = "Error"
 dialog_autowrap = true
 
 [node name="SceneTransitionCover" parent="." instance=ExtResource( 6 )]
+
```
Poobslag, line 146: Revert this.

Poobslag: Reverted.

PR #945, opened 2021-06-24
Added missing level creature_ids. if_condition -> available_if.

Poobslag:
Replaced 'if_condition' with 'available_if'. The term if_condition is woefully
ambiguous when there are multiple conditional statements.

Added missing creature_id entries in worlds.json. This should allow
more levels to be available when walking around on the overworld.

### DIFF
--- a/project/src/main/ui/level-select/level-library.gd
+++ b/project/src/main/ui/level-select/level-library.gd
@@ -85,9 +85,13 @@ func first_unfinished_level_id_for_creature(creature_id: String) -> String:
 		var world_lock: WorldLock = world_lock_obj
 		for level_id in world_lock.level_ids:
 			var level_lock: LevelLock = _level_locks[level_id]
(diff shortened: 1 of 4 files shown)


project/src/main/ui/level-select/level-library.gd
```diff
@@ -85,9 +85,13 @@ func first_unfinished_level_id_for_creature(creature_id: String) -> String:
...
-			if level_lock.creature_id == creature_id and not level_lock.is_locked() \
-					and not PlayerData.level_history.is_level_finished(level_lock.level_id):
-				return level_lock.level_id
+			if level_lock.creature_id != creature_id:
+				continue
+			if level_lock.is_locked():
+				continue
+			if PlayerData.level_history.is_level_finished(level_lock.level_id):
+				continue
+            return level_lock.level_id
```
Poobslag, line 94: Replace these spaces with tabs.

Poobslag: Replaced

PR #946, opened 2021-06-24
LevelLibrary tests, fixes. BoolExpression renames.

Poobslag:
Added unit test for LevelLibrary.

LevelLibrary method renames. Changed is_locked -> is_level_locked to
disambiguate it from a potential (but nonexistent) is_world_locked function.
Changed first_unfinished_level_id_for_creature -> next_creature_level for
conciseness.

Renamed 'ChatBool' classes to 'BoolExpression'. These have utility
in selecting levels, too. They're not just for chats.

Added missing 'locked_until' condition for goodbye_bones.


project/src/main/ui/bool-expression-parser.gd
```diff
@@ -132,10 +132,10 @@ func _init(string: String, subject = null) -> void:
 
 
 """
-Parses a list of ChatTokens into a tree of ChatExpressions which can be evaluated.
+Parses a list of ChatTokens into a tree of BoolExpressions which can be evaluated.
```
Poobslag, line 135: These are not 'ChatTokens', they are 'BoolTokens'.

Poobslag: Changed to `BoolTokens`.

PR #948, opened 2021-06-28
Chatscript improvements: link moods, multiline metadata, conditions

Poobslag:
Chat links can now have their own moods. Sometimes we want multiple chat
choices to link to the same item, but to have different moods.

Chatscript parser now accepts multiple meta items on different lines. This
feels more natural, especially with lengthy meta items or meta items
ending with punctuation.

ChatTree now supports 'say_if' metadata item. This allows a cutscene's
dialog to change based on the game state.

ChatTree now supports 'creature_mood' metadata item. This allows a single line
of dialog to provoke reactions from multiple creatures at once.

Renamed bool-expression tests, which retained the outdated 'chat-bool' name
when the class was renamed.


project/src/main/ui/chat/chat-tree.gd
```diff
@@ -104,9 +104,27 @@ func advance(link_index := -1) -> bool:
 		# advance through the current chat branch
 		_position.index += 1
 		did_increment = true
+	skip_unmet_conditions()
 	return did_increment
 
 
+func skip_unmet_conditions() -> void:
```
Poobslag, line 111: Add a method-level comment

project/src/main/ui/chat/chat-ui.gd
```diff
@@ -137,14 +137,16 @@ func _on_ChatFrame_all_text_shown() -> void:
...
+			if chat_event.link_moods[i] != -1:
+				mood = chat_event.link_moods[i]
+			elif _chat_advancer.chat_tree.events.has(link):
 				var first_event: ChatEvent = _chat_advancer.chat_tree.events[link][0]
 				if first_event and first_event.mood:
-					moods.append(first_event.mood)
-				else: moods.append(-1)
-			else:
-				moods.append(-1)
+					mood = first_event.mood
```
Poobslag, line 148: Comment this conditional so it's clear what the branches mean

project/src/main/ui/chat/chat-tree.gd
Poobslag: Added comment, `Skip any dialog lines whose ...`

project/src/main/ui/chat/chat-ui.gd
Poobslag: Added comments, `use the mood from the chat link...`

PR #956, opened 2021-07-01
Smoother cutscene cameras. Better camera adjustments for fat creatures.

Poobslag:
Creatures entering or exiting a cutscene now launch a tween which smoothly moves
the camera instead of lerping. The previous lerp effect caused the camera to
leap into motion which was unpleasant.

Cutscene camera zooms and pans slower than before. The rapid zooming and
panning was also unpleasant.

Better camera adjustments for fat creatures. Camera is adjusted vertically to
keep fat creature's faces in frame. Camera is zoomed out when fat creatures are
in a cutscene, this logic was accidentally removed in f0f31d1

hello-everyone-000 cutscene no longer fades in three creatures on the
first line of dialog. This was unnoticable, but resulted in the camera
(correctly) zooming in on just the two visible creatures, and pulling out to all
five creatures, which looked strange.

### DIFF
--- a/project/src/main/world/overworld-camera.gd
+++ b/project/src/main/world/overworld-camera.gd
@@ -9,27 +9,19 @@ const AUTO_ZOOM_OUT_DISTANCE := 100.0
 # amount of empty space around characters
 const CAMERA_BOUNDARY := 160
 
-# amount of time spent zooming in and out
-const ZOOM_DURATION := 1.0
+# duration of sweeping pan/zoom during cutscenes and conversations
+const ZOOM_DURATION := 1.8
 
-# maximum zoom amount for conversations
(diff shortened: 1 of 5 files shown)


project/src/main/world/overworld-camera.gd
```diff
@@ -82,8 +57,88 @@ func set_close_up(new_close_up: bool) -> void:
 		return
 	
 	close_up = new_close_up
+	_tween_camera_to_chatters()
+
+
+"""
+Returns a rectangle representing the area which should be captured by the camera.
+
+This includes all of the creatures in the conversation. It
```
Poobslag, line 66: Finish this comment.

Poobslag: Removed second unfinished sentence, it seemed unnecessary.

PR #961, opened 2021-07-06
Sort signal connections. Reset creature colors, emote positions

Poobslag:
delint.sh now sorts signal connections. By default Godot shuffles signal
connections randomly, and sometimes randomly drops signal connections as
well. This makes it easy to accidentally remove signals, and for those
deletions to be obfuscated in a huge sea of changes. Alphabetizing the
list should give us a chance to catch those kinds of bugs during code
review.

The 'reset_creature' option in CreatureVisuals.tscn now resets their
colors, shader parameters, and emote positions. This cuts down on
unnecessary noise during commits involving the CreatureVisuals class.

Delint.sh now uses bash script instead of shell script. In POSIX sh,
proecss substitution is undefined.

### DIFF
--- a/project/src/main/world/creature/emote-player.gd
+++ b/project/src/main/world/creature/emote-player.gd
@@ -394,12 +394,26 @@ green.
 """
 func _post_unemote() -> void:
 	for emote_sprite in _emote_sprites:
-		emote_sprite.scale = Vector2(2.0, 2.0)
-		emote_sprite.rotation_degrees = 0.0
+		emote_sprite.frame = 0
 		emote_sprite.modulate = Color.transparent
+		emote_sprite.rotation_degrees = 0.0
+		emote_sprite.scale = Vector2(2.0, 2.0)
+		emote_sprite.position = Vector2(0, 0)
+		if emote_sprite.material:
+			if emote_sprite.material.get("blend_mode"):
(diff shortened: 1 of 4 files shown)


todo.txt
```diff
@@ -0,0 +1 @@
+-'error calling method from signal dna_loaded'?
```
Poobslag, line 1: remove this 'todo' file

project/src/main/world/creature/emote-player.gd
```diff
@@ -394,12 +394,23 @@ green.
...
-		emote_sprite.rotation_degrees = 0.0
+		emote_sprite.frame = 0
 		emote_sprite.modulate = Color.transparent
+		emote_sprite.rotation_degrees = 0.0
+		emote_sprite.scale = Vector2(2.0, 2.0)
+		emote_sprite.position = Vector2(0, 0)
+		if emote_sprite.material:
+			if emote_sprite.material.get("blend_mode"):
+				emote_sprite.material.blend_mode = SpatialMaterial.BLEND_MODE_MIX
+			if emote_sprite.material.has_method("set_shader_param") and Engine.is_editor_hint():
```
Poobslag, line 405: Leave a comment describing why we only reset this in the editor.

Poobslag: Added a comment, `In the editor, we reset shader parameters to known values ...`

todo.txt

PR #962, opened 2021-07-07
Sorted .tscn signals. Improved sorting logic.

Poobslag:
Changed signal sort order. The old order was signal, from, to, method.
The new order is from, signal, to, method. This matches and extends
Godot's default sort order to avoid breaking the game's current
behavior.

Removed cruft from CreatureVisuals.tscn - positions, shader parameters
and blend modes which frequently changed when manipulating creatures in
the editor.


project/src/main/ui/menu/PracticeMenu.tscn
```diff
@@ -73,9 +73,9 @@ __meta__ = {
 }
 
 [node name="Buttons" type="HBoxContainer" parent="VBoxContainer/Mode"]
-margin_left = 92.0
+margin_left = 88.0
```
Poobslag, line 76: Why did these values change?

Poobslag: These values are dictated by a LayoutManager, and it's possible the LayoutManager's behavior changed in a recent version of Godot.

PR #963, opened 2021-07-07
Refactor: '!' -> 'not'. LevelSelect nodepath. ChatLibrary 'state' dict.

Poobslag:
Replaced '!' operator with 'not'. GDScript guidelines recommend plain english
versions of operators.

level-select.gd now references its LevelButtons instance with a NodePath. This
requires more code but prevents things from breaking when reorganizing
the node tree.

Removed vestigial 'creature_chat_state' from ChatLibrary. The 'notable'
and 'level_id' functionality were moved into other places, so we were just
passing around a dictionary with a creature_id.

Minor tweaks to ChatLibrary to ensure focused_chattable_chat_tree() is
not invoked when starting a level. This prevents creatures from making
incongruent chit chat before their level is launched.

Changed CurrentLevel.push_cutscene_trail() to push_preroll_trail() to
make its purpose more clear.

Fixed boatricia's chat tree to use 'available_if'. The absence of this
flag was skipping her filler dialog.

### DIFF
--- a/project/src/main/ui/chat/chat-library.gd
+++ b/project/src/main/ui/chat/chat-library.gd
@@ -12,10 +12,12 @@ const PREROLL_SUFFIX := "000"
 const POSTROLL_SUFFIX := "100"
 
 """
-Loads a conversation for the specified creature.
+Returns the chat tree for the specified creature.
 
 Each creature has a sequence of conversations defined by their 'chat selectors'. This method goes through a creature's
 chat selectors until it finds one suitable for the current game state.
+
+Returns null if the chat tree cannot be found.
 
 Parameters:
 	'creature': The creature whose conversation should be returned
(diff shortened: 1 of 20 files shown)


project/src/main/ui/chat/chat-library.gd
```diff
@@ -23,14 +23,8 @@ Parameters:
...
-	var level_id: String = state["level_id"]
-	var chat_tree := chat_tree_for_creature_def(creature.creature_def, state)
-
-	if level_id:
-		# schedule a level to launch when the chat completes
-		CurrentLevel.set_launched_level(level_id)
+func chat_tree_for_creature(creature: Creature) -> ChatTree:
+	var chat_tree := chat_tree_for_creature_def(creature.creature_def)
 
 	return chat_tree
```
Poobslag, line 29: Unnecessary local before return.

project/src/test/ui/chat/test-chat-library.gd
```diff
@@ -52,28 +51,28 @@ func before_each() -> void:
 func test_greeting() -> void:
 	PlayerData.chat_history.delete_history_item("chat/gurus750/greeting_001")
 	
-	assert_eq(ChatLibrary.select_from_chat_selectors(_chat_selectors, _state, _filler_ids), "greeting_001")
+	assert_eq(ChatLibrary.select_from_chat_selectors(_chat_selectors, "gurus750", _filler_ids), "greeting_001")
```
Poobslag, line 54: Extract a utility method in this test to avoid copy/pasting the userid everywhere.

project/src/main/ui/chat/chat-library.gd
Poobslag: Inlined unused `chat_tree_for_creature_def` function.

project/src/test/ui/chat/test-chat-library.gd
Poobslag: Extracted `_select_from_chat_selectors()` utility method.

PR #965, opened 2021-07-08
Fix unnecessary type hints, ': Array = []'

Poobslag:


project/src/main/world/creature/creature-loader.gd
```diff
@@ -154,7 +154,7 @@ Returns a CreatureCurve for drawing belly colors for a type of body.
 """
 func new_belly(body_key: String, body_colors_key: String) -> CreatureCurve:
 	var result: CreatureCurve
-	var key: String = "%s %s" % [body_key, body_colors_key]
+	var key := "%s %s" % [body_key, body_colors_key]
```
Poobslag, line 157: Does this type hint work here? It doesn't elsewhere

Poobslag: It works as long as you use array notation `"foo" % [bar]`. It doesn't work without it `"foo" % bar`

PR #969, opened 2021-07-09
Cutscene decision refactoring. wall-library rename.

Poobslag:
Extracted common cutscene decision logic. Replaced copy/paste
'has_preroll' check with implicit null check

Renamed wall-library to indoor-obstacle-library. It now includes things
like countertops and sinks, more than just walls.


project/src/main/puzzle/level/current-level.gd
```diff
@@ -96,34 +96,48 @@ func switch_level(new_settings: LevelSettings) -> void:
...
+	'chat_tree': The cutscene to evaluate.
 	
-	var chat_tree: ChatTree
-	if level_id and ChatLibrary.has_preroll(level_id):
-		chat_tree = ChatLibrary.chat_tree_for_preroll(level_id)
-
-	if chat_tree and (chat_tree.location_id or force):
-		SceneTransition.push_trail(chat_tree.cutscene_scene_path())
+	'ignore_player_preferences': If 'true', the player's preferences will be ignored, and the method will only
+		evaluate if the player's seen the cutscene and whether its 'skip_if' condition is met.
```
Poobslag, line 110: rephrase "evaluate if" to "check whether"

Poobslag: Rephrased

PR #977, opened 2021-07-16
Replaced ambiguous 'chat' prefix with 'creature', 'level'

Poobslag:
Before, the 'chat' prefix could be used to point to resources in
/assets/main/creatures/primary or /assets/main/chat. These are now
referenced by different prefixes, 'creature' and 'chat', and there is a
third path 'level' for levels.

Replace hyphens with underscores in creature library. Creature library
used hyphens for '#fatness-023#' and a long time ago, for a creature
with an id of 'i-n-cognito'


project/assets/test/turbofat-24cc.json
```diff
@@ -101,5 +101,258 @@
...
+        "rhonk": 1,
+        "mewow": 1,
+        "pairet": 1.1,
+        "tunathy": 1.69334,
+        "gordo": 1,
+        "terpion": 1.207477,
+        "sandrew": 1.887856,
+        "i-n-cognito": 1.3
+      }
+    }
```
Poobslag, line 356: This section should be in the new turbofat-2743 save file, and begs the question of how on earth the current tests passed

Poobslag: Moved to `turbofat-2743.save`. The tests were passing because CreatureLibrary.reset() did not erase fatness data, so the data was carried forward from the 24cc test. I've fixed CreatureLibrary.reset().

PR #979, opened 2021-07-21
Creature spawners, obstacle spawners

Poobslag:
Replaced specific 'MarshCrystalForSign' scene with generic obstacle
spawners. Obstacle spawners accept an obstacle scene, properties and node
groups. If a condition is met, they will create the specified scene with
the specified properties. This is similar to the 'MarshCrystalForSign'
but should be generic enough to work for a wide variety of obstacles.

Similarly, added creature spawners for spawning creatures. This handles
some creature-specific features like moving shadows, scenario-specific chat
icons and stool shadows.

Moved Sensei earlier in Overworld scene tree. They're important so it doesn't
make sense to bury them with other NPCs.


project/src/main/world/obstacle-spawner.gd
```diff
@@ -0,0 +1,59 @@
...
+# the PackedScene of the spawned obstacle
+export (PackedScene) var TargetPackedScene: PackedScene
+
+# the properties of the spawned obstacle
+export (Dictionary) var target_properties: Dictionary
+
+# the groups of the spawned obstacle
+export (Array) var target_groups: Array
+
+onready var _overworld_world: OverworldWorld = get_node(overworld_world_path)
```
Poobslag, line 22: This onready variable should go below the exported variables.

project/src/main/world/creature-spawner.gd
```diff
@@ -0,0 +1,82 @@
...
+var CreaturePackedScene: PackedScene = load("res://src/main/world/creature/Creature.tscn")
+
+# stool the spawned creature sits on, if any
+var _stool: Stool
+
+onready var _overworld_world: OverworldWorld = get_node(overworld_world_path)
+
+func _ready() -> void:
+	if stool_path:
+		_stool = get_node(stool_path)
```
Poobslag, line 36: Does get_node(null) return null, with no error? If so, `_stool` can be an onready variable.

project/src/main/world/spawn.gd
```diff
@@ -1,18 +1,35 @@
...
 export (String) var id: String
 
+export (float) var elevation: float
+
+# stool the creature sits on, if any
+var _stool: Stool
+
+func _ready() -> void:
+	if stool_path:
+		_stool = get_node(stool_path)
```
Poobslag, line 23: Verify `get_node` functionality -- maybe move this as an onready variable.

project/src/main/world/creature-spawner.gd
Poobslag: get_node(null) returns an error -- this logic is necessary. There is a `get_node_or_null` but that will conceal misconfigured/nonexistent node paths.

project/src/main/world/spawn.gd
Poobslag: get_node(null) returns an error -- this logic is necessary.

project/src/main/world/obstacle-spawner.gd
Poobslag: Moved.

PR #980, opened 2021-07-21
Cutscene queuing. Creatures don't reorient for thoughts.

Poobslag:
Cutscene queueing logic. Chatscript now supports 'next scene foo' to
queue up a cutscene.

Creatures don't face eachother during thoughts. This avoids an odd
circumstance where you think 'i shouldn't disturb this customer' but they turn
to face you anyways.

### DIFF
--- a/project/src/main/world/creature/creature.gd
+++ b/project/src/main/world/creature/creature.gd
@@ -151,6 +151,17 @@ func _physics_process(delta: float) -> void:
 	_maybe_play_bonk_sound(old_non_iso_velocity)
 
 
+"""
+Increases the collision shape size for fatter creatures.
+
+This is not done on all creatures, because having fatter creatures collide with chairs/tables differently as they gain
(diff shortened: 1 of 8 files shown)


project/src/main/world/creature/creature.gd
```diff
@@ -151,6 +151,15 @@ func _physics_process(delta: float) -> void:
 	_maybe_play_bonk_sound(old_non_iso_velocity)
 
 
+"""
+Increases the collision shape size for fatter creatures.
+
+This is not done on all creatures, because having fatter creatures collide with chairs/tables differently as they gain weight introduces problems. It's easier to leave their collision shapes consistent and just have a few visual bugs here and there.
```
Poobslag, line 157: This line is much longer than 120 characters. Why did the delint script not catch this?

Poobslag: The 'long lines' clause in the delinter was not running properly. I've fixed it and shortened this line.

PR #982, opened 2021-07-21
More buttercup cutscenes. Customers arrive/leave, plants grow.

Poobslag:
New cutscenes, 'buttercup cafe is having trouble', 'soliciting buttercup
customers'. These make use of the new features to have creatures spawn
in seats, and to have cutscenes follow one after another.

Diota cuts down a lot of marsh crystals as the storyline progresses, and
new crystals sprout near the Turbo Fat restaurant.

Customers leave the Buttercup Cafe as the storyline progresses, and new
customers arrive at the Turbo Fat restaurant.


project/assets/main/creatures/primary/diota/filler-000.chat
```diff
@@ -1,3 +1,3 @@
 {"version": "2476", "filler": true}
 
-diota: -_- Oh hey! You're here again. ...Yep. Bye!
+diota: >_< Are you going to order something?
```
Poobslag, line 3: Worth documenting the rationale behind changing this filler.

Poobslag: Added comment, `Her first filler conversation was...`

PR #987, opened 2021-07-28
Multiple save slots. Split SystemData from PlayerData.

Poobslag:
Save slots are accessible through the Settings/Misc menu. Selecting a
new save slot prompts the user and redirects them to the splash page.
Save slots can also be deleted, the player is prompted twice.

Split SystemData/SystemSave from PlayerData/PlayerSave.

With multiple save slots, we need a centralized place to store which
save slot is active. It makes sense to store configuration information
like language and keybinds to this location. Players don't want to rebind their
keys and reconfigure their language every time they start a new game.

### DIFF
--- a/project/src/test/test-old-player-save.gd
+++ b/project/src/test/test-old-player-save.gd
@@ -4,9 +4,11 @@ Tests backwards compatibility with older save formats.
 """
 
 const TEMP_FILENAME := "test-ground-lucky.save"
+const TEMP_LEGACY_FILENAME := "test-snatch-lucky.save"
 
 func before_each() -> void:
 	PlayerSave.data_filename = "user://%s" % TEMP_FILENAME
+	PlayerSave.legacy_filename = "user://%s" % TEMP_LEGACY_FILENAME
 	
 	# don't increment playtime during this test or it ruins our assertions
 	PlayerData.seconds_played_timer.stop()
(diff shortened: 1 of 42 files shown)


project/src/test/test-old-player-save.gd
```diff
@@ -115,22 +118,26 @@ func test_2743() -> void:
...
 	# chat counts should use underscores, and new prefixes
 	assert_eq(PlayerData.chat_history.chat_counts.get("creature/richie"), 6)
 	assert_eq(PlayerData.chat_history.chat_counts.get("creature/shirts"), 2)
-	
+
 	# filler counts should use underscores, and new prefixes
 	assert_eq(PlayerData.chat_history.filler_counts.get("creature/richie"), 6)
 	assert_eq(PlayerData.chat_history.filler_counts.get("creature/shirts"), 2)
-	
+
```
Poobslag, line 129: Revert these whitespace changes

project/src/test/test-system-save.gd
```diff
@@ -0,0 +1,45 @@
...
+
+
+func test_save_and_load() -> void:
+	SystemData.volume_settings.set_bus_volume_linear(VolumeSettings.MASTER, 0.841)
+	SystemData.volume_settings.set_bus_volume_linear(VolumeSettings.MUSIC, 0.695)
+	SystemData.volume_settings.set_bus_volume_linear(VolumeSettings.SOUND, 0.279)
+	SystemData.volume_settings.set_bus_volume_linear(VolumeSettings.VOICE, 0.405)
+	PlayerSave.save_player_data()
+	PlayerData.reset()
+	PlayerSave.load_player_data()
```
Poobslag, line 41: This test needs to save/reset/load system data -- not player data

project/src/test/test-old-player-save.gd
Poobslag: Reverted.

project/src/test/test-system-save.gd
Poobslag: Fixed. Verified tests still pass.

PR #997, opened 2021-08-10
Fixed ChatUiDemo. Chatscript comments.

Poobslag:
ChatUiDemo's chat files were never converted from json to chatscript, so
ChatUiDemo threw errors instead of demonstrating the UI. I've converted
its chat files.

I removed chat-blank. Chat-blank included chat lines which can't be represented
in chatscript, and which resulted in strange things happening in the chat UI
anyways. I don't think we need to support null/blank chat lines.

Chatscript comments are now '//' to avoid conflict with '#player#'


project/assets/test/ui/chat/chat-full.chat
```diff
@@ -3,4 +3,5 @@
 boatricia: ^_^ Oh! I remember you from before. I'm Boatricia, I'm sort of new here and trying to make some friends.
 player: Nice to meet you! People here call me #player#.
 player: /._. Well! ...If you're trying to make some friends, I'll try to cook you something special okay? Then maybe you can be my friend.
+// Comment giving this chat some context
```
Poobslag, line 6: Consider a less opaque comment like `// Example comment`. The phrase 'comment giving this chat...' can be interpreted as a request or a plan to finish something later.

Poobslag: Replaced with `example comment`

PR #1000, opened 2021-08-13
Bugs: CreatureSpawner chat animations, chattables without chat extents

Poobslag:
Fixed bug where CreatureSpawner didn't register chatters, causing their
chat animations to not work

Fixed bug where ChatUi.get_chatter_bounding_box threw errors for chattables
without chat extents

### DIFF
--- a/project/src/main/world/chattable-manager.gd
+++ b/project/src/main/world/chattable-manager.gd
@@ -75,7 +75,7 @@ func refresh_creatures() -> void:
 	_creatures_by_id.clear()
 	for creature_obj in get_tree().get_nodes_in_group("creatures"):
 		var creature: Creature = creature_obj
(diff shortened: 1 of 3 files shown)


project/src/main/world/chattable-manager.gd
```diff
@@ -201,20 +201,20 @@ func focused_chattable_creature_id() -> String:
 
 
 """
-Remove the specified creature from the '_creatures_by_id' mapping.
+Add the specified creature to the '_creatures_by_id' mapping.
```
Poobslag, line 204: Grammatically, these comments should say 'adds', 'removes' and not 'add', 'remove'.

Poobslag: Changed to adds/removes.

PR #1001, opened 2021-08-13
Improved left/right nametag algorithm

Poobslag:
Instead of basing it on which way the creature is facing, it now
calculates the midpoint of the conversation and attaches metadata to
each ChatEvent specifying whether the nametag should appear on the left
or right.

This is more consistent. Creatures might talk to each other when they're
both walking in the same direction. Or a creature might turn left and
right during a cutscene.

### DIFF
--- a/project/src/main/ui/chat/chat-frame.gd
+++ b/project/src/main/ui/chat/chat-frame.gd
@@ -53,7 +53,7 @@ Animates the chat UI to gradually reveal the specified text, mimicking speech.
 Also updates the chat UI's appearance based on the amount of text being displayed and the specified color and
 background texture.
 """
-func play_chat_event(chat_event: ChatEvent, nametag_right: bool, squished: bool) -> void:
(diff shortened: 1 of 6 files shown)


project/src/main/ui/chat/chat-frame.gd
```diff
@@ -53,7 +53,7 @@ Animates the chat UI to gradually reveal the specified text, mimicking speech.
 Also updates the chat UI's appearance based on the amount of text being displayed and the specified color and
 background texture.
 """
-func play_chat_event(chat_event: ChatEvent, nametag_right: bool, squished: bool) -> void:
+func play_chat_event(chat_event: ChatEvent, squished: bool) -> void:
```
Poobslag, line 56: Where is this method reference? Shouldn't its invocations change too?

Poobslag: Updated method references in ChatFrameDemo, ChatUi

PR #1003, opened 2021-08-13
Added prologue which introduces the player to story mode.

Poobslag:
The first time the player picks story mode, they're shown a prologue
which shows Fat Sensei and the player walking towards Merrymellow Marsh
alongside some narration and dialog.

Creatures no longer try to emote with their hands while walking. This
was causing them to have three or more hands visible at once.

The overworld now forcibly plays chill music. This change was necessary
because the prologue uses tutorial music, but we didn't want this persisting
through the overworld.

Added support for new 'destination' chatscript tag. This allows
cutscenes to conclude somewhere other than where they take place.

Relocated player and sensei start position, and made Richie spawn
indoors before the first level. Hopefully this helps players figure out
where everything is and find the levels they need to find.


project/src/main/world/overworld-walk-ui.gd
```diff
@@ -0,0 +1,24 @@
+extends OverworldUi
```
Poobslag, line 1: Add class, method level comments.

project/src/main/world/overworld-walk-ui.gd
```diff
@@ -0,0 +1,24 @@
...
+
+const WALK_REACTION_DELAY := 0.4
+
+func _apply_chat_event_meta(chat_event: ChatEvent, meta_item: String) -> void:
```
Poobslag, line 5: This should call the superclass's method, right?

project/src/main/world/overworld-walk-ui.gd
Poobslag: Added comments, `UI elements for the overworld walking scene...` and others

project/src/main/world/overworld-walk-ui.gd
Poobslag: Added `._apply_chat_event_meta(...` call

PR #1015, opened 2021-08-20
Added level variants where lines don't drop. Added experimental world.

Poobslag:
These two new level variants are grouped in an 'Experimental' grouping
of levels. This is unlocked from the beginning, so players can test them
easily.


project/src/main/puzzle/level/blocks-during-rules.gd
```diff
@@ -3,9 +3,30 @@ class_name BlocksDuringRules
...
+const LINE_CLEAR_TYPES_BY_NAME := {
+	"default": LineClearType.DEFAULT,
+	"float": LineClearType.FLOAT,
+	"float_fall": LineClearType.FLOAT_FALL,
+}
+
 # if true, the entire playfield is cleared when the player tops out
 var clear_on_top_out := false
 
+var line_clear_type: int = LineClearType.DEFAULT
```
Poobslag, line 23: Comment this field. Make sure to document it in the wiki as well.

Poobslag: Added comment, `whether lines drop following a line clear`. Updated wiki.

PR #1018, opened 2021-08-24
Added level timers. Added level which gradually inserts rows.

Poobslag:
Levels can now define timers. Timers can trigger events like rotating a
piece or inserting a line.

### DIFF
--- a/project/src/main/puzzle/level-timer-manager.gd
+++ b/project/src/main/puzzle/level-timer-manager.gd
@@ -1,0 +1,51 @@
+extends Node
+"""
+Starts and stops level-specific timers during a puzzle.
+
+Some levels define behavior where something happens every 5 seconds. These timer definitions are kept in the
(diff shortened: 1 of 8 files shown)


project/src/main/puzzle/level-timer-manager.gd
```diff
@@ -0,0 +1,51 @@
+extends Node
+"""
+Starts and stops level-specific timers during a puzzle.
+
+Some levels define behavior where something happens every 5 seconds. These timer definitions are kept in the
+LevelTimers script, but the actual timer objects which count down are kept here.
+"""
+
+# The maximum amount of timers which a level can define
+var MAX_TIMER_COUNT := 1
```
Poobslag, line 10: This should be a constant.

Poobslag: Changed to `const`.

PR #1026, opened 2021-08-31
Extracted SaveConverter from OldPlayerSave, OldSystemSave

Poobslag:
All save conversion logic was doing the same logic for iterating over
save data items, converting each one, and updating the version number.
Additionally, there was redundancy where multiple places had the concept
of 'these 12 versions can be upgraded', both in the transformation logic
and the 'is old save item' checking logic.

These redundancies have been resolved by extracting a new SaveConverter
class. This class can be configured with methods to convert specific
versions of save items.

Fixed some OldPlayerSave tests, which would error (and pass!) if it
tried to assert something about an empty array

### DIFF
--- a/project/src/main/save-converter.gd
+++ b/project/src/main/save-converter.gd
@@ -1,0 +1,103 @@
+class_name SaveConverter
+"""
+Provides backwards compatibility with older save formats.
+
+SaveConverter can update the 'version' tag, but any other version-specific updates must be defined externally. These
+version-specific updates can be incorporated via SaveConverter's 'add_method' method.
+"""
+
+"""
+An externally defined method which provides version-specific updates.
+"""
+class ConversionMethod:
+	var object: Object # The object containing the method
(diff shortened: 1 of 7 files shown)


project/src/main/save-converter.gd
```diff
@@ -0,0 +1,98 @@
...
+"""
+Adds a new externally defined method which provides version-specific updates.
+
+SaveConverter does not have logic for converting specific save data versions. This conversion logic must be defined on
+an external object and incorporated via this 'add_method' method.
+
+Parameters:
+	'object': The object containing the method
+	
+	'method': The name of the method which performs the conversion
```
Poobslag, line 32: The comment should describe more details about the signature and contract of this method.

Poobslag: Expanded comment, `This method should accept a SaveData object and return a modified SaveData object.`

PR #1030, opened 2021-09-02
Refactoring: music fading, overworld_world_path

Poobslag:
Many different scripts had logic saying 'if a chill song isn't playing, fade in
a new chill song'. This logic is now centralized within MusicPlayer with a
'fade_in' parameter.

Added default overworld_world_path for spawners. This prevents us from needing
to assign it everywhere, cutting down on cruft in overworld scenes

### DIFF
--- a/project/src/main/music/music-transition.gd
+++ b/project/src/main/music/music-transition.gd
@@ -13,10 +13,6 @@ func _ready() -> void:
 	SceneTransition.connect("fade_in_started", self, "_on_SceneTransition_fade_in_started")
 
 
-func is_fading_in_bgm() -> bool:
-	return true if _faded_bgm else false
-
-
 """
 When the scene fades back in, we un-fade any music we previously faded out.
 """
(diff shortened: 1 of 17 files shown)


project/src/main/music/music-transition.gd
```diff
@@ -13,10 +13,6 @@ func _ready() -> void:
 	SceneTransition.connect("fade_in_started", self, "_on_SceneTransition_fade_in_started")
 
 
-func is_fading_in_bgm() -> bool:
-	return true if _faded_bgm else false
-
-
```
Poobslag, line 19: Does MusicTransition still work OK?

project/src/main/ui/menu/practice-menu.gd
```diff
@@ -68,6 +68,7 @@ var _rank_lowlights := []
 
 func _ready() -> void:
 	ResourceCache.substitute_singletons()
+	MusicPlayer.play_chill_bgm()
```
Poobslag, line 71: Should this also be added to the tutorial menu?

Poobslag: I've added this logic to `tutorial-menu.gd`

project/src/main/music/music-transition.gd
Poobslag: MusicTransition broke as a result of these changes. I've fixed it.

PR #1042, opened 2021-09-15
Row insert data. Moved 'blocks_start' to 'tiles/start'.

Poobslag:
LineInsert rules can now reference a specific group of tiles to insert.
These tiles are inserted from top to bottom. They're included in the
new 'LevelTiles' class within LevelSettings.

Moved 'blocks_start' data under the new 'tiles' JSON entry. The idea
of separating 'blocks_start' data from 'blocks_during' data was nice,
but it didn't work because of how the data is organized. Collections of tiles
are saves as dictionaries. Collections of rules are saved as string arrays.

Incremented level version from 19c5 to 297a.


project/src/main/puzzle/004.json
```diff
@@ -0,0 +1,182 @@
+{
+  "version": "297a",
```
Poobslag, line 2: Remove this file.

project/src/test/puzzle/level/test-level-trigger.gd
```diff
@@ -46,19 +46,19 @@ func test_after_line_cleared_0456() -> void:
 
 
 func test_phase_definition_params_unkeyed() -> void:
```
Poobslag, line 48: These method names should change.

project/src/main/puzzle/004.json
Poobslag: Removed.

project/src/test/puzzle/level/test-level-trigger.gd
Poobslag: Renamed to `test_dict_config_*`

PR #1044, opened 2021-09-17
Pickups before level start. Redundant LevelTiles fields.

Poobslag:
Pickups now appear before the player hits 'start'. StarSeeds are not
shown before the player hits 'start' because their locations are random,
but pickups are deterministic.

Removed redundant block_cells and pickup_cells LevelTiles fields. These
might make sense to restore if the order of the cells matters, but for
now i don't see a reason to keep them.

Added experiment 004 to worlds.json.

Improved failure messages for some LevelSettings tests.

### DIFF
--- a/project/assets/main/puzzle/levels/experiment/004.json
+++ b/project/assets/main/puzzle/levels/experiment/004.json
@@ -1,6 +1,8 @@
 {
   "version": "297a",
   "start_speed": "0",
+  "title": "A Little Extra",
+  "description": "Some fun little bonuses at the bottom. But you have to work for it!",
   "finish_condition": {
     "type": "score",
     "value": "400"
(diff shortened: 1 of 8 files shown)


project/assets/main/puzzle/levels/experiment/004.json
```diff
@@ -1,6 +1,8 @@
 {
   "version": "297a",
   "start_speed": "0",
+  "title": "A Little Extra",
+  "description": "Some fun little bonuses at the bottom... But you have to work for them!.",
```
Poobslag, line 5: Extra punctuation

Poobslag: Fixed punctuation

PR #1045, opened 2021-09-17
Level editor now allows editing of multiple tilesets.

Poobslag:
Extracted PlayfieldEditor from LevelEditor


project/src/main/editor/puzzle/playfield-editor-control.gd
```diff
@@ -0,0 +1,97 @@
+class_name PlayfieldEditorControl
+extends HBoxContainer
```
Poobslag, line 2: Needs comments.

Poobslag: Added comments throughout, `UI control for editing the playfield....`

PR #1046, opened 2021-09-21
Pickup improvements: renames, pickups at start

Poobslag:
Renamed pickup methods to match other similar classes.

"set_pickup" is a better name than "add" because the pickups replace other
pickups in the same spot, and it is similar to "set_cell"

"clear" matches other similar classes

"remove" is better as a public method. Nobody uses it yet but its
intuitive behavior could be used in a demo or on a specific level.

Pickups are now spawned in the _ready() method so they will appear
when the puzzle scene first starts up.

### DIFF
--- a/project/src/main/puzzle/pickups.gd
+++ b/project/src/main/puzzle/pickups.gd
@@ -40,6 +40,47 @@ func _ready() -> void:
 	PuzzleState.connect("game_prepared", self, "_on_PuzzleState_game_prepared")
 	CurrentLevel.connect("settings_changed", self, "_on_Level_settings_changed")
 	Pauser.connect("paused_changed", self, "_on_Pauser_paused_changed")
+	_prepare_pickups_for_level()
+
+
+"""
+Adds or replaces a pickup in a playfield cell.
+"""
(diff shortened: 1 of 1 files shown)


project/src/main/puzzle/pickups.gd
```diff
@@ -69,28 +70,29 @@ func _prepare_pickups_for_level() -> void:
...
-	
-	var pickup: Pickup = SeedPickupScene.instance()
-	pickup.food_type = _food_type_for_cell(box_type, cell)
-
-	pickup.position = _puzzle_tile_map.map_to_world(cell + Vector2(0, -3))
-	pickup.position += _puzzle_tile_map.cell_size * Vector2(0.5, 0.5)
-	pickup.position *= _puzzle_tile_map.scale
-	pickup.scale = _puzzle_tile_map.scale
-	pickup.z_index = 4 # in front of the active piece
+func set_pickup(cell: Vector2, box_type: int) -> void:
```
Poobslag, line 81: These methods are out of order; public methods should precede private ones

Poobslag: Reordered methods.

PR #1066, opened 2021-09-27
Removed instant das squish kludge, I don't like it.

Poobslag:
This kludge was introduced in 69672802 to prevent pieces from slipping past
nooks during squish moves. I can't remember the exact circumstances where this
was frustrating, but activating DAS results in weird piece handling at times.
Performing a squish move and moving the piece into an open area makes the piece
suddenly zip across the playfield. If this 'slipping past nooks' issue is a
consistent problem we need a more specific fix for it -- perhaps something that
only activates for a frame, only activates at 20G, or something like that.


Poobslag:
I also experimented with "rocket drop squish moves" as the original PR suggests. You can't slide pieces into nooks during rocket drops, but you can't slide pieces into nooks during hard drops either. This seems consistent.

PR #1067, opened 2021-09-27
Fixed visual bug when food flies to a customer who didn't order it.

### LINKED ISSUE #876: If customers change rapidly, food-flying logic can do weird things
If the player feeds a customer a lot of food, and then breaks their combo twice in rapid succession, a new customer will replace the original customer who was fed. However, the food will continue flying toward their mouth, and will not be eaten.

Poobslag:
If a food is flying to a customer who's been replaced with the current customer,
the food now disappears. This is preferable to the previous behavior where the
food would fly to the current customer's mouth, but they wouldn't eat it.

Closes #876.

### DIFF
--- a/project/src/main/puzzle/food-item.gd
+++ b/project/src/main/puzzle/food-item.gd
@@ -79,9 +79,13 @@ func _physics_process(delta: float) -> void:
 			_source_pos += velocity * delta
 			_target_pos = _get_target_pos.call_funcv(_target_pos_arg_array)
 			
-			# The y coordinate changes at a constant rate while the x coordinate follows a parabolic path
-			position.y = lerp(_source_pos.y, _target_pos.y, _flight_percent)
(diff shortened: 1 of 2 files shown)


project/src/main/puzzle/food-item.gd
```diff
@@ -79,9 +79,12 @@ func _physics_process(delta: float) -> void:
 			_source_pos += velocity * delta
 			_target_pos = _get_target_pos.call_funcv(_target_pos_arg_array)
 			
-			# The y coordinate changes at a constant rate while the x coordinate follows a parabolic path
-			position.y = lerp(_source_pos.y, _target_pos.y, _flight_percent)
-			position.x = lerp(_source_pos.x, _target_pos.x, 2.1 * pow(_flight_percent, 1.6) - 1.1 * _flight_percent)
+			if _target_pos == Vector2.INF:
+				queue_free()
```
Poobslag, line 83: This is worth a comment, it's very surprising.

Poobslag: Added a comment, `get_target_pos returns Vector2.INF if the food should be removed`

PR #1070, opened 2021-09-27
Inserted boxes aren't autocleared. random_tiles_start flag.

Poobslag:
Inserted boxes aren't automatically cleared upon finishing a level.

Added random_tiles_start flag for levels which want more randomness.


project/src/main/puzzle/level/blocks-during-rules.gd
```diff
@@ -40,6 +40,8 @@ var line_clear_type: int = LineClearType.DEFAULT
 # whether pickups move with the playfield blocks
 var pickup_type: int = PickupType.DEFAULT
 
+var random_tiles_start: bool = false
```
Poobslag, line 43: comment

Poobslag: `whether inserted rows should ...`

PR #1079, opened 2021-10-01
LevelSettings can be written as json.

Poobslag:
Expanded RuleParser to handle writing rules as json. This involved
creating an extensible framework which can handle different property
types. It can handle basic bools/ints/floats/strings which are used
throughout multiple classes, but can also handle things like the piece
lists used in PieceTypeRules and the ShowRank properties used in RankRules.

This also required changing many other classes to be serializable to
json such as milestones, triggers and phase conditions.

Boolean properties can now be specified using json strings like
'clear_on_finish false' for consistency with non-boolean properties.

Added Utils.to_bool() utility as a workaround for Godot #27529
(https://github.com/godotengine/godot/issues/27529).

### DIFF
--- a/project/src/main/puzzle/level/milestone.gd
+++ b/project/src/main/puzzle/level/milestone.gd
@@ -49,3 +49,26 @@ func from_json_dict(json: Dictionary) -> void:
 	for key in json.keys():
 		if not key in ["type", "value"]:
 			set_meta(key, json.get(key))
+
+
+func to_json_dict() -> Dictionary:
+	var result := {
+		"type": Utils.enum_to_snake_case(MilestoneType, type),
+		"value": value,
+	}
+	for meta_key in get_meta_list():
+		result[meta_key] = get_meta(meta_key)
+	return result
+
+
+"""
+Returns 'true' if all of the milestone's properties are currently set to their default values.
+
+Milestones whose properties are set to their default values are omitted from our json output.
+"""
+func is_default() -> bool:
+	var result := true
+	result = result && (type == MilestoneType.NONE)
+	result = result && (value == 0)
+	result = result && (get_meta_list().empty())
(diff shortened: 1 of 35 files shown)


project/src/main/puzzle/level/milestone.gd
```diff
@@ -49,3 +49,21 @@ func from_json_dict(json: Dictionary) -> void:
...
+	var result := {
+		"type": Utils.enum_to_snake_case(MilestoneType, type),
+		"value": value,
+	}
+	for meta_key in get_meta_list():
+		result[meta_key] = get_meta(meta_key)
+	return result
+
+
+func is_default() -> bool:
```
Poobslag, line 64: This is worth a comment.

project/src/main/puzzle/level/rule-parser.gd
```diff
@@ -1,51 +1,427 @@
...
+func to_json_array() -> Array:
+	var result := []
+	for property_parser in _property_parsers:
+		if not property_parser.is_default():
+			result.append_array(property_parser.to_json_strings())
+	return result
+
+
+"""
+Returns 'true' if all of the rule's properties are currently set to their default values.
```
Poobslag, line 419: It might be good to hint at why we do this.

project/src/main/puzzle/level/milestone.gd
Poobslag: Added comment: `Milestones whose properties are set to their default values are omitted from our json output.`

project/src/main/puzzle/level/rule-parser.gd
Poobslag: Espanded comment, `Rules whose properties are set to their default values are omitted from our json output.`

PR #1082, opened 2021-10-02
Fixed bug where inserted lines did not work.

Poobslag:
This stems from a type conversion error which affects released versions of the
game. When performing float operations, Godot will haphazardly store the result
as a float or an int depending on the circumstances. In the Godot editor, it
likes storing them as ints more often. But when exported, it likes storing them
as floats more often.


Poobslag:
This bug stems from [Godot #41336](https://github.com/godotengine/godot/issues/41336). TileMap is inherently designed to work with both float values (via methods like `get_cellv`) and int values (via methods like `get_cell`). 99% of the time these play well together, but if you are not careful you can end up referencing `used_cell.y` under the assumption that it is an int, when it is actually a float -- resulting in unpredictable results.

Worse yet, these variable types can be different when the game is exported with debug mode on or off.

PR #1084, opened 2021-10-04
Enforce player creature_id to be PLAYER_ID.

Amavect:
Let's have set_player_def always set the creature_id to
PLAYER_ID so the player's creature_id can never be out of sync.
This makes the creature_id field in the json save file unneeded
and makes some extra assignments unnecessary.


Poobslag:
To be honest, I wrote out a four or five paragraph explanation rationalizing why this change was dangerous or could break things. ...But I hammered on your code a bit and this is a great change.

Thanks so much for your contribution!

PR #1091, opened 2021-10-11
Improved V piece climb kick.

Poobslag:
A V piece will now climb up a one-tile stairwell instead of being pushed
off to the left.

A twitch streamer 'PrincessMomo13' attempted this kick and was surprised
the piece wouldn't climb in this way, and I agreed it was surprising.


project/src/test/puzzle/piece/test-piece-kicks-v.gd
```diff
@@ -390,14 +408,14 @@ func test_lwallkick_ccw2() -> void:
 		"     ",
 		"     ",
 		"vvv  ",
-		"  v  ",
+		": v  ",
```
Poobslag, line 411: This behavior is fine but it shouldn't be called "test_lwallkick"

PR #1095, opened 2021-10-14
Allow piece movement during post squish frames.

Poobslag:
This allows pieces to squeeze into gaps after a squish move.
Specifically, for squish moves like this:

```

  oo
  oo
:::
:
:
::
::
::
```

Before, this setup would result in the O piece dropping before it could
be slid to the left.


project/src/main/puzzle/piece/PieceManager.tscn
```diff
@@ -157,6 +157,7 @@ piece_mover_path = NodePath("../Mover")
 [node name="Squisher" type="Node" parent="Physics"]
 script = ExtResource( 29 )
 input_path = NodePath("../../Input")
+piece_mover_path = NodePath("../Mover")
```
Poobslag, line 160: This is not needed.

PR #1114, opened 2021-11-03
Unused constants. top_out_penalty float. push_warning.

Poobslag:
Removed unused constants from blocks-during-rules script.

Changed RankRules.top_out_penalty from int to float for consistency with
RankRules.success_bonus

Replaced push_error with push_warning in non-fatal places. For recoverable
errors, like a missing locale or a background that can't scroll, push_warning
is more appropriate. push_error should be for things which prevent the game
from being playable.

### DIFF
--- a/project/src/main/puzzle/level/blocks-during-rules.gd
+++ b/project/src/main/puzzle/level/blocks-during-rules.gd
@@ -15,22 +15,6 @@ enum PickupType {
 	FLOAT_REGEN, # pickups float in place and regenerate if collected
 }
 
-# key: (string) line clear type which appears in level definitions
-# value: (int) an enum from LineClearType
-const LINE_CLEAR_TYPES_BY_NAME := {
-	"default": LineClearType.DEFAULT,
-	"float": LineClearType.FLOAT,
-	"float_fall": LineClearType.FLOAT_FALL,
-}
-
-# key: (string) pickup type which appears in level definitions
-# value: (int) an enum from PickupType
-const PICKUP_TYPES_BY_NAME := {
-	"default": PickupType.DEFAULT,
-	"float": PickupType.FLOAT,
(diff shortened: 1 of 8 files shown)


project/src/main/editor/puzzle/playfield-editor-control.gd
```diff
@@ -82,7 +82,7 @@ func _on_PlayfieldNav_add_tiles_key_pressed() -> void:
 			break
 	
 	if not new_key:
-		push_error("Couldn't add tiles key: Too many conflicts")
+		push_warning("Couldn't add tiles key: Too many conflicts")
```
Poobslag, line 85: These feel better suited as errors; you are pressing a button, and the button has no effect

project/src/main/world/cutscene-manager.gd
```diff
@@ -86,7 +86,7 @@ Transitions to the cutscene at the front of the queue, staying at the current le
 """
 func replace_cutscene_trail() -> void:
 	if not is_front_chat_tree():
-		push_error("CutsceneManager._queue.front (%s) is not a cutscene" % [_queue.front()])
+		push_warning("CutsceneManager._queue.front (%s) is not a cutscene" % [_queue.front()])
```
Poobslag, line 89: This feels better suited as an error. Not loading a cutscene is very bad and could make the game softlock.

Poobslag: (Both of the errors in this file.)

project/src/main/editor/puzzle/playfield-editor-control.gd
Poobslag: Reverted.

project/src/main/world/cutscene-manager.gd
Poobslag: Reverted.

PR #1115, opened 2021-11-03
Merged 'Seed' and 'Star' scenes into 'StarSeed'

Poobslag:
Having these two nodes as children of the same parent simplifies
the 'star-seeds.gd' script which only needs to know about a single
scene. It also eliminates the need for a 'base_scale' and
'base_rotation' property of the children.


project/src/main/puzzle/wobbler.gd
```diff
@@ -16,10 +16,6 @@ export (float) var avg_pulse_amount := 0.2
 # How many seconds the sprite should take to shrink to minimum/maximum scale
 export (float) var avg_pulse_period := 3.0
 
-# The unmodified scale/rotation before pulsing/spinning
-export (Vector2) var base_scale := Vector2(1.0, 1.0) setget set_base_scale
-var base_rotation := 0.0 setget set_base_rotation
-
 # The food type represented by this wobbler
 var food_type := 0
```
Poobslag, line 20: Is this wobbler's food_type property still referenced?

project/src/main/puzzle/star-seed.gd
```diff
@@ -0,0 +1,27 @@
+class_name StarSeed
+extends Node2D
```
Poobslag, line 2: Add a class-level comment

project/src/main/puzzle/wobbler.gd
Poobslag: No; I've removed the food_type property.

project/src/main/puzzle/star-seed.gd
Poobslag: Added comment: "Draws a shadowy star or seed inside a snack box or cake box."

PR #1116, opened 2021-11-04
Grade changes: pieces per minute, endurance/speed

Poobslag:
Grade system now awards an 'M' rank for speed as long as you clear 65
lines per minute. It used to only award an 'M' rank based on the maximum
theoretical speed for each level, which was in some cases much faster
than anybody can play (yet).

Grade system now separates endurance/speed and grades them separately.
This means that clearing 90/100 lines for a Marathon course scores a
low grade, but playing at 90% speed for a Sprint course scores a high
grade.

Adjusted Rank courses to provide a more linear increase in difficulty.
There used to be a monumental jump between Rank 6d/7d which required you to
play at about 50 pieces per minute, and Rank 8d/9d which required double
that. It now gradually ramps up from 50 ppm to 110 ppm.

### DIFF
--- a/project/src/main/puzzle/rank-calculator.gd
+++ b/project/src/main/puzzle/rank-calculator.gd
@@ -10,8 +10,8 @@ const BEST_RANK := RankResult.BEST_RANK
 # These RDF (rank difference factor) constants from (0.0 - 1.0) affect how far apart the ranks are. A number like 0.99
 # means the ranks are really narrow, and you can fall from rank 10 to rank 20 with only a minor mistake. A number like
 # 0.96 means the ranks are more forgiving.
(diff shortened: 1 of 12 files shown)


project/src/main/puzzle/rank-calculator.gd
```diff
@@ -137,10 +144,12 @@ static func min_frames_per_line(piece_speed: PieceSpeed) -> float:
 """
 Calculates the lines per minute for a master player.
```
Poobslag, line 145: Fix this inaccurate comment.

Poobslag: Changed to `...for a specific rank`

PR #1128, opened 2021-11-18
Added toggle setting for 'Soft Drop Lock Cancel'

Poobslag:
Twitch user PrincessMomo13 couldn't get used to the 'soft drop cancels
piece lock' functionality of Turbo Fat, so I've made it optional. It's
enabled by default.

### DIFF
--- a/project/src/main/puzzle/piece/states/prelock.gd
+++ b/project/src/main/puzzle/piece/states/prelock.gd
@@ -3,7 +3,11 @@ extends State
 
 func update(piece_manager: PieceManager) -> String:
 	var new_state_name := ""
-	piece_manager.move_piece()
+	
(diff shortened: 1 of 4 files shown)


project/src/main/puzzle/piece/piece-physics.gd
```diff
@@ -58,7 +58,7 @@ func move_piece(piece: ActivePiece) -> bool:
 		result = true
 		if piece.lock > 0:
 			if dropper.did_hard_drop:
-				# hard drop doesn't cause lock reset
+				# hard drop doesn't cause lock cancel
```
Poobslag, line 61: Revert this; they're referring to a lock reset which is different from a lock cancel

Poobslag: Reverted

PR #1130, opened 2021-11-19
Added regions for career mode. Travel distance increases with rank.

Poobslag:
Regions are defined in the 'career-worlds.json' file. Each region
occupies a range of distance, and has its own piece speeds and levels.

Travelled distance increases with rank. The better you do, the further
you go.


project/src/test/Tests.tscn
```diff
@@ -13,6 +13,6 @@ __meta__ = {
 _run_on_load = true
 _should_maximize = true
 _yield_between_tests = false
-_file_prefix = "test-"
+_file_prefix = "test-rank-calculator"
```
Poobslag, line 16: Revert this.

Poobslag: Reverted.

PR #1157, opened 2021-12-08
Career mode now shows an overworld scene.

Poobslag:
The overworld scene is mostly empty and only includes the player and
sensei. But eventually, it will include customers and scenery.


project/src/main/world/career-map.gd
```diff
@@ -46,15 +32,13 @@ var _piece_speed: String
 
 var _duration_calculator := DurationCalculator.new()
```
Poobslag, line 33: Is duration_calculator still used?

Poobslag: Yes, it's used to determine the height of the level button.

PR #1161, opened 2021-12-10
Customers emote when selected. Results screen line breaks.

### LINKED ISSUE #1061: End of level results display sometimes wraps its lines
image

image

This seems rare -- whatever calculation we're performing to ensure the window is the correct width might just sometimes be off by a pixel or two.

Poobslag:
Customers in career mode now emote when you choose their level. Usually
they're happy but sometimes they're anxious.

Decreased the number of periods in the end of level results display, to
prevent it from wrapping its lines unnecessarily.

Closes #1061.

### DIFF
--- a/project/src/main/world/career-world.gd
+++ b/project/src/main/world/career-world.gd
@@ -10,7 +10,16 @@ const X_DIST_BETWEEN_CUSTOMERS := 200
 ## vertical distance separating the customers from the player's path
 const Y_DIST_BETWEEN_CUSTOMERS_AND_PATH := 80
 
+## List of moods customers have when their level is chosen.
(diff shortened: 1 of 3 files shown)


project/src/main/world/career-world.gd
```diff
@@ -10,10 +10,32 @@ const X_DIST_BETWEEN_CUSTOMERS := 200
...
+	ChatEvent.Mood.SIGH0,
+	ChatEvent.Mood.SWEAT0,
+	ChatEvent.Mood.THINK0,
+]
+
 export (NodePath) var player_path2d_path: NodePath
 
 ## List of Customer instances for the level's customers. Index 0 holds the leftmost customer.
 var customers := []
+var customer_moods := []
```
Poobslag, line 38: This should be private and have a comment.

Poobslag: Made private and added a comment, `List of moods each customer has when their level is chosen`

PR #1168, opened 2021-12-14
Career mode decreases distance if you pick certain levels.

### LINKED ISSUE #1140: Career Mode should decrease distance if you pick certain levels
Currently Career Mode lets you pick freely between three levels, so players are never incentivized to play their least favorite level and -- in fact, will almost always just play a level they're comfortable with

The two unpicked levels should carry a mild penalty (1, 2 distance) so that players have a more interesting choice, rather than always picking their favorite

Poobslag:
The rightmost level is selected by default. Selecting an earlier level
reduces the player's distance travelled by up to 4 steps.

Closes #1140.

### DIFF
--- a/project/src/main/career-data.gd
+++ b/project/src/main/career-data.gd
@@ -132,6 +132,37 @@ func is_boss_level() -> bool:
 	return result
 
 
+## Returns the player's distance penalties for picking different levels.
+##
+## Upon beating a level, the player is advanced by distance_earned. If they select one of the two leftmost levels,
+## a penalty is applied and they don't travel as far.
+##
+## Returns:
+## 	An array of positive ints corresponding to the penalty for selecting each level. Index 0 corresponds to the
+## 	earliest level.
+func distance_penalties() -> Array:
+	var result := [0, 0, 0]
+	
+	if is_boss_level():
+		# boss levels only have one choice, there is no penalty
+		return result
+	
+	# adjust result[0]
+	if distance_earned < 0: result [0] = 1 # small penalty for level selection after failing a boss stage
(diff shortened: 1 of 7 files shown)


project/src/main/ui/career/career-distance-label.gd
```diff
@@ -3,28 +3,43 @@ extends Control
...
+export (NodePath) var level_select_path: NodePath
+
 onready var _label := $Label
 
+## UI components for career mode's level select buttons
+onready var _level_select: CareerLevelSelect = get_node(level_select_path)
+
 func _ready() -> void:
 	PlayerData.career.connect("distance_travelled_changed", self, "_on_CareerData_distance_travelled_changed")
+
```
Poobslag, line 15: remove this whitespace

project/src/main/career-data.gd
```diff
@@ -132,6 +132,33 @@ func is_boss_level() -> bool:
 	return result
 
 
+## Returns the player's distance penalties for picking different levels.
+##
+## Upon beating a level, the player is advanced by distance_earned. If they select one of the two leftmost levels,
+## a penalty is applied and they don't travel as far.
```
Poobslag, line 138: document the data type, and the fact that they're positive

project/src/main/ui/career/career-distance-label.gd
Poobslag: Removed

project/src/main/career-data.gd
Poobslag: Expanded comment: `An array of positive ints corresponding...`

PR #1186, opened 2021-12-20
Added 'Give Up' button to career mode

Poobslag:
Settings menu now supports two quit buttons, so a settings menu can have
both 'Save + Quit' and 'Give Up' for career mode.

Note: Settings menu's HBoxContainers/VBoxContainers must have a
separation of 0. Having a non-zero value causes the buttons at the
bottom to have strange choppy fonts.


project/src/main/ui/settings/SettingsMenu.tscn
```diff
@@ -66,7 +66,6 @@ pause_mode = 2
 script = ExtResource( 5 )
 
 [node name="Bg" type="ColorRect" parent="."]
-visible = false
```
Poobslag, line 69: Revert this.

Poobslag: Reverted

PR #1195, opened 2021-12-22
Added 'shuffle_inserted_lines' level feature.

Poobslag:
Replaced 'random_tiles_start' feature with 'shuffle_inserted_lines
slice'. shuffle_inserted_lines also supports 'bag' to bag shuffle the
lines, and may eventually support other shuffles such as a dice shuffle,
or a subtler less random shuffle.


project/src/main/puzzle/level/blocks-during-rules.gd
```diff
@@ -1,6 +1,12 @@
 class_name BlocksDuringRules
 ## Blocks/boxes/pickups which appear or disappear while the game is going on.
 
+enum ShuffleInsertedLinesType {
+	NONE,
```
Poobslag, line 5: add a line comment for this

Poobslag: Added line comment, `do not shuffle`

PR #1199, opened 2021-12-23
More career levels. Career levels are in phone menu.

Poobslag:
Added new 'Bottomless Pit', 'Creeping Crevice' levels to career mode.

Career levels are now available through the phone menu for practice. They've
replaced the '/levels/experiment/' levels.

Reduced combo factor for levels with a combo factor over 1.00. A combo
factor of 1.2 requires 21 points per line -- a value which is impossible to
reach no matter how efficiently you play.


project/assets/main/puzzle/career-worlds.json
```diff
@@ -37,7 +37,7 @@
           "id": "career/snack_box_blitz"
         },
         {
-          "id": "career/a_little_garbage"
+          "id": "career/cookie_crevice"
```
Poobslag, line 40: What about bottomless pit, bottomless pit 2, and back to back?

project/assets/test/ui/level-select/career-worlds.json
```diff
@@ -25,7 +25,7 @@
           "id": "career/veggie_haters"
         },
         {
-          "id": "career/muffin_impossible"
+          "id": "career/bottomless_pit"
```
Poobslag, line 28: Revert

project/assets/main/puzzle/career-worlds.json
Poobslag: Added bottomless pit, bottomless pit 2, back to back

project/assets/test/ui/level-select/career-worlds.json
Poobslag: Reverted

PR #1220, opened 2022-01-05
Refactoring: SettingsMenuBottom, CareerWin subcomponents

Poobslag:
Extracted child nodes within SettingsMenu, CareerWin. Both these scripts had
some very long node paths for logic which could be handled by the bottom node.
I've moved the node paths and logic.

### DIFF
--- a/project/src/main/ui/career/career-win.gd
+++ b/project/src/main/ui/career/career-win.gd
@@ -4,82 +4,13 @@ extends Control
 ## This summary screen includes things like how much money the player earned and how many customers they served, as
 ## well as a visual map showing their progress through the world.
 
+# Daily step thresholds to trigger positive feedback.
+const DAILY_STEPS_GOOD := 25
+const DAILY_STEPS_OK := 8
+
 export (NodePath) var obstacle_manager_path: NodePath
 
-## Negative titles shown when the player fails, travelling zero steps during a career session.
-var _bad_titles := [
-	tr("OH DEAR WHAT HAPPENED"),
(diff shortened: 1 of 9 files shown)


project/project.godot
```diff
@@ -75,6 +75,11 @@ _global_script_classes=[ {
 "path": "res://src/main/ui/level-select/career-region.gd"
 }, {
 "base": "Control",
+"class": "CareerWin",
```
Poobslag, line 78: Revert this. CareerWin does not need to be a named class.

project/src/main/resource-cache.gd
```diff
@@ -117,18 +117,18 @@ func substitute_singletons() -> void:
...
-			
 			# add the singleton value
-			parent.add_child(_singletons[non_singleton.name])
-			parent.move_child(_singletons[non_singleton.name], wallpaper_index)
+			parent.add_child(_singletons[non_singleton_name])
+			parent.move_child(_singletons[non_singleton_name], wallpaper_index)
 		else:
 			# we have no singleton value stored; store a new singleton value
-			_singletons[non_singleton.name] = non_singleton
+			_singletons[non_singleton_name] = non_singleton
```
Poobslag, line 131: Revert this.

project/project.godot
Poobslag: Reverted.

project/src/main/resource-cache.gd
Poobslag: Reverted.

PR #1233, opened 2022-01-11
Career mode keeps level buttons focused

Poobslag:
Previously in career mode, when clicking the settings button and then clicking
'OK', the level buttons would be deselected. It now reselects the previously
selected level button.

Previously in career mode, when clicking to choose a new start level,
the level buttons would be deselected. It now reselects the previously
selected level button.

### DIFF
--- a/project/src/main/ui/career/career-map-level-select.gd
+++ b/project/src/main/ui/career/career-map-level-select.gd
@@ -7,6 +7,7 @@ signal level_button_focused(button_index)
 export (PackedScene) var LevelSelectButtonScene: PackedScene
 
 var _duration_calculator := DurationCalculator.new()
+var _prev_focused_level_button_index := -1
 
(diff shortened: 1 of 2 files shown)


project/src/main/ui/career/career-map-level-select.gd
```diff
@@ -7,6 +7,7 @@ signal level_button_focused(button_index)
 export (PackedScene) var LevelSelectButtonScene: PackedScene
 
 var _duration_calculator := DurationCalculator.new()
+var _prev_focused_level_button_index := -1
```
Poobslag, line 10: Comment this field.

Poobslag: The field name is self-documenting; a comment like `The previously focused level button index` feels redundant and I can't think of anything to add.

PR #1239, opened 2022-01-12
Fixed many tutorial bugs.

Poobslag:
Fixed bug where basics tutorial (and other tutorials) couldn't be
finished. This bug was introduced in 6a1f692e by making
PuzzleState.reset() disable the game_active flag. Having this flag
disabled in tutorials meant that they wouldn't end the level, because
they didn't think the player was playing.

Fixed bug where mouse/keyboard input was ignored after quitting to the
tutorial menu. This is because the TutorialMenu scene was missing its
scene transition cover, which is responsible for restoring input.

Fixed bug where tutorial buttons like "Ok, I get it!" weren't responding
to mouse input. This occurred because the PuzzleMessages control was
blocking the mouse signals. I've disabled mouse input for the
PuzzleMessages control (not for its children.)

Fixed bug where squish tutorial would change levels before messages were
shown completely. This is because it was calling the global
'change_scene' method instead of the method specific to tutorials.

Cake tutorial launches upbeat music at end, like the other tutorials.

### DIFF
--- a/project/src/main/puzzle/tutorial/tutorial-module.gd
+++ b/project/src/main/puzzle/tutorial/tutorial-module.gd
@@ -33,6 +33,18 @@ func start_customer_countdown() -> void:
 	puzzle.start_level_countdown()
 
 
+## Resets the timer and scores and dismisses the sensei, so the player can serve some real customers.
+##
+## Parameters:
+## 	'messages': Array of string messages to be shown when the sensei is dismissed.
+func dismiss_sensei(messages: Array) -> void:
+	PuzzleState.reset()
+	PuzzleState.game_active = true
+	puzzle.scroll_to_new_creature()
+	hud.set_messages(messages)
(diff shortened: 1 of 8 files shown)


project/src/main/puzzle/tutorial/tutorial-module.gd
```diff
@@ -33,6 +33,15 @@ func start_customer_countdown() -> void:
 	puzzle.start_level_countdown()
 
 
+## Resets the timer and scores and dismisses the sensei, so the player can serve some real customers.
+func dismiss_sensei(messages: Array) -> void:
```
Poobslag, line 37: Document this parameter.

Poobslag: Added comment, `Array of string messages to be shown ...`

PR #1259, opened 2022-01-31
Merrymellow Marsh prologue plays in career mode

Poobslag:


project/src/main/ui/level-select/career-region.gd
```diff
@@ -1,6 +1,8 @@
 class_name CareerRegion
 ## Stores information about a block of levels for career mode.
 
+const PROLOGUE_CHAT_KEY_NAME := "prologue"
```
Poobslag, line 4: This is worth a comment

PR #1288, opened 2022-02-10
Replaced specific cutscene/story scenes with swappable ones.

Poobslag:
Deleted OverworldWalk, OverworldIndoors. Instead of loading these
scenes, we now load generic Cutscene and Overworld scenes with a
different swappable environment.

Added a new swappable Cutscene scene. This scene is used by the CutsceneDemo,
story mode and career mode. It is similar to the overworld scene where it has
an environment and logic for initializing creatures.

The new cutscene-specific scene, world, camera, and creature spawning
logic is borrowed from bits and pieces of the old Overworld and
OverworldWalk scenes.

The generic overworld no longer plays cutscenes, so the logic for its
non-interactive cutscene mode has been removed.


project/src/main/world/CareerMap.tscn
```diff
@@ -51,7 +49,7 @@ action = "ui_cancel"
 [sub_resource type="ShortCut" id=14]
 shortcut = SubResource( 22 )
 
-[node name="Node" type="Node"]
+[node name="Node2" type="Node"]
```
Poobslag, line 52: `Node2`?

project/src/main/world/cutscene-ui.gd
```diff
@@ -0,0 +1,87 @@
+extends OverworldUi
+## UI elements for the overworld walking scene.
```
Poobslag, line 2: This comment is out of date.

project/src/main/world/cutscene-camera.gd
```diff
@@ -1,18 +1,37 @@
+class_name CutsceneCamera
 extends Camera2D
 ## Overworld camera for the walking scene. Follows the two characters.
```
Poobslag, line 3: This comment is inaccurate.

project/src/main/world/cutscene-ui.gd
Poobslag: Changed to `UI elements for overworld cutscenes.`

project/src/main/world/cutscene-camera.gd
Poobslag: Changed to `Camera for overworld cutscenes.`.

PR #1298, opened 2022-02-17
Refactorings: Career mode advancement logic improvements.

Poobslag:
Fixed bug where CareerData referenced 'PlayerData.career' instead of referencing
itself.

Broke up CareerData.advance_clock() function which was growing large.

Extracted new CareerData.process_puzzle_result() function from Puzzle to keep the
career-specific logic separate from the puzzle logic.

Extracted new level_choice_count() method. Currently this only toggles
based on whether the player is playing a boss level, but in the future
there are other reasons why the player will have only one level to
choose.

### DIFF
--- a/project/src/main/career-data.gd
+++ b/project/src/main/career-data.gd
@@ -198,7 +198,7 @@ func is_career_cutscene() -> bool:
 
 ## Returns 'true' if the player has completed the boss level in the specified region
 func is_region_cleared(region: CareerRegion) -> bool:
-	return PlayerData.career.max_distance_travelled > region.distance + region.length - 1
+	return max_distance_travelled > region.distance + region.length - 1
 
 
 ## Returns 'true' if the current career mode distance corresponds to an uncleared boss level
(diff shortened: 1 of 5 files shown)


project/src/main/career-data.gd
```diff
@@ -331,6 +320,55 @@ func set_hours_passed(new_hours_passed: int) -> void:
...
+##
+## If the player skips or fails a level, this has consequences such as advancing the clock.
+func process_puzzle_result() -> void:
+	if not PuzzleState.game_ended:
+		# player skipped a level
+		advance_clock(0, false)
+		skipped_previous_level = true
+
+
+## Apply some of the player's distance earned, either advancing the player or throwing it away.
```
Poobslag, line 333: apply -> applies

project/src/main/ui/career/career-distance-label.gd
```diff
@@ -24,13 +24,15 @@ func _refresh_label() -> void:
 	# Display the distance travelled with the distance penalty applied
 	_label.text = StringUtils.comma_sep(PlayerData.career.distance_travelled - distance_penalty)
 	
-	# Append the distance_option with the distance penalty applied
-	if PlayerData.career.distance_earned > 0:
+	# Append the distance_option with the distance penalty, banked steps applied
```
Poobslag, line 27: remove 'banked steps' reference

Poobslag: Removed.

PR #1305, opened 2022-02-18
'bossio' cheat improvements.

Poobslag:
'bossio' cheat now searches more diligently for a region with a boss
level. If none can be found, a 'cheat deactivated' sound plays.

'bossio' cheat now marks cutscenes as unviewed, and the boss level as
unplayed. This is slightly more future proof, just in case our logic
eventually changes to key off of whether or not the boss level has been
beaten. (It's currently based on the maximum distance travelled)


project/src/main/ui/career/career-map-ui.gd
```diff
@@ -69,9 +101,10 @@ func _on_SettingsMenu_other_quit_pressed() -> void:
...
-		_force_cutscene()
-		detector.play_cheat_sound(true)
-	elif cheat == "bossio":
-		_force_boss_level()
-		detector.play_cheat_sound(true)
+	match cheat:
+		"cutsio":
+			_force_cutscene()
+			detector.play_cheat_sound(true)
+		"bossio":
```
Poobslag, line 108: Alphabetize these cheats

Poobslag: Alphabetized

PR #1306, opened 2022-02-18
CareerCutsceneLibrary improvements: traversal order, special keys

Poobslag:
Fixed tree traversal order with
CareerCutsceneLibrary.filter_chat_key_pairs(). It was appending items to
the start and end of a queue, which worked for simple linear tree
traversal cases but failed for more complex trees.

CareerCutsceneLibrary now ignores special keys such as 'prologue' and
'intro_level'. Before, these keys would receive the same treatment as
numeric cutscenes like '10-a'. The only reason they were never selected
is because the player views them immediately upon entering a region, and
we avoid showing the player duplicate cutscenes. But this would cause
noticable bugs for boss or epilogue cutscenes.

Renamed CareerCutsceneLibrary methods. The initialization method is now
called 'refresh_chat_key_pairs' to disambiguate it from
'find_chat_key_pairs' which is responsible for searching the cache for
specific nodes.

### DIFF
--- a/project/src/main/ui/chat/career-cutscene-library.gd
+++ b/project/src/main/ui/chat/career-cutscene-library.gd
@@ -34,6 +34,16 @@ extends Node
 ## Default resource path containing career cutscenes.
 const DEFAULT_CAREER_CUTSCENE_ROOT_PATH := "res://assets/main/chat/career"
 
+## List of specially named chat keys which should be excluded when browsing for cutscenes to play
+const SPECIAL_CHAT_KEY_NAMES := [
+	"prologue",
+	"intro_level",
+	"intro_level_end",
+	"boss_level",
+	"boss_level_end",
+	"epilogue",
+]
+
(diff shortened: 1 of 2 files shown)


project/src/main/ui/chat/career-cutscene-library.gd
```diff
@@ -136,7 +146,7 @@ func set_all_chat_key_pairs(new_all_chat_key_pairs: Array) -> void:
 ## Returns:
 ## 	A set of chat keys for cutscenes the player has already seen. This includes the leaf nodes for cutscenes which
 ## 	have been seen, and branch nodes where all child cutscenes have been seen.
-func _exhausted_chat_keys(chat_key_roots: Array) -> Dictionary:
+func exhausted_chat_keys(chat_key_roots: Array) -> Dictionary:
```
Poobslag, line 149: This method does not need to be public (yet)

Poobslag: Reverted.

PR #1307, opened 2022-02-18
Added epilogue cutscene to career mode. Added 'epilio' cheat.

Poobslag:
Merrymellow Marsh defines an epilogue cutscene. This epilogue plays after
the last normal cutscene of the region. When activated, the player is
automatically advanced past the area.

Added new 'epilio' cheat to force an epilogue to play.

### DIFF
--- a/project/src/main/ui/career/career-map-ui.gd
+++ b/project/src/main/ui/career/career-map-ui.gd
@@ -61,20 +61,73 @@ func _find_region_with_boss_level() -> CareerRegion:
 func _force_boss_level() -> bool:
 	var new_region := _find_region_with_boss_level()
 	
+	# set their max_distance_travelled so that the boss level isn't skipped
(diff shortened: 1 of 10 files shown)


project/src/main/ui/career/career-map-ui.gd
```diff
@@ -61,20 +61,74 @@ func _find_region_with_boss_level() -> CareerRegion:
...
 	var new_region := _find_region_with_boss_level()
 	
+	# set their max_distance_travelled so that the boss level isn't skipped
+	PlayerData.career.max_distance_travelled = PlayerData.career.distance_travelled
+	
+	# reload the CareerMap scene
+	SceneTransition.change_scene()
+
+
+
```
Poobslag, line 71: Too much whitespace.

Poobslag: Fixed

PR #1316, opened 2022-02-25
Career mode ensures cutscenes and levels match.

Poobslag:
Chatscript now allows cutscenes to specify chefs/customers so they can
be matched up between cutscenes and levels.

When selecting levels in career mode, we now only allow the player to choose
levels featuring creatures from potential upcoming cutscenes.

When selecting a cutscene in career mode, we now only select cutscenes featuring
creatures from the chosen level.

### DIFF
--- a/project/src/main/ui/level-select/career-level-library.gd
+++ b/project/src/main/ui/level-select/career-level-library.gd
@@ -114,3 +114,72 @@ func _load_raw_json_data() -> void:
 			regions[i].length = regions[i + 1].distance - regions[i].distance
 		else:
 			regions[i].length = CareerData.MAX_DISTANCE_TRAVELLED
+
+
+## Returns a collection of chefs and customers required for potential upcoming cutscenes.
+##
+## Based on the upcoming cutscenes, we sometimes restrict the player's level choice to one or two levels where those
+## cutscenes make sense. This method returns the chefs/customers required for the upcoming cutscenes. As long as one
+## of these chefs/customers is selected by the player, there will be at least one appropriate cutscene which features
+## them.
+##
+## Returns:
+##	A dictionary with two entries:
+## 		'chef_ids' lists ids for creatures who acts as the chef for a cutscene.
+##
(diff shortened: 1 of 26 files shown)


project/src/main/ui/level-select/career-level.gd
```diff
@@ -1,6 +1,8 @@
 class_name CareerLevel
 ## Stores career-mode-specific information about a level.
 
+const ANONYMOUS_CUSTOMER := "#anonymous_customer#"
```
Poobslag, line 4: Document this constant.

project/src/main/ui/level-select/career-level-library.gd
```diff
@@ -114,3 +114,72 @@ func _load_raw_json_data() -> void:
...
+					result["chef_ids"].append(chat_tree.chef_id)
+			elif chat_tree.customer_ids:
+				# If a cutscene specifies both a chef AND customer, we ignore the customer. The chef is the only
+				# creature who appears on the map.
+				if not chat_tree.customer_ids[0] in result["customer_ids"]:
+					result["customer_ids"].append(chat_tree.customer_ids[0])
+			else:
+				if not CareerLevel.ANONYMOUS_CUSTOMER in result["customer_ids"]:
+					result["customer_ids"].append(CareerLevel.ANONYMOUS_CUSTOMER)
+	return result
```
Poobslag, line 152: We could use a chef_ids and customer_ids field which are assembled at the end. This would be a little less error prone (no fat-fingering magic strings)

project/src/main/ui/level-select/career-level.gd
Poobslag: Added comment, `Unique customer ID assigned to levels...`

project/src/main/ui/level-select/career-level-library.gd
Poobslag: Added `chef_ids`, `customer_ids` variables which are assembled at the end.

PR #1320, opened 2022-03-01
Cutscene renames: LevelHistory.add(), CutsceneManager, others

Poobslag:
Renamed CutsceneManager to CutsceneQueue. Its behavior is really just
'here's a bunch of cutscenes to play before and after a level', so this
name is more specific.

Improved symmetry between LevelHistory's methods and ChatHistory's
methods. Simple methods names like add() and delete() are slightly
ambiguous, since this could involve deleting a single result, or
deleting an entire set of results. In fact, the methods were
inconsistent and delete DID delete an entire set of results. The names
now communicate that.


project/src/main/world/Cutscene.tscn
```diff
@@ -1,13 +1,11 @@
-[gd_scene load_steps=9 format=2]
```
Poobslag, line 1: Revert this.

project/src/test/Tests.tscn
```diff
@@ -7,9 +7,6 @@ margin_right = 740.0
 margin_bottom = 250.0
 rect_min_size = Vector2( 740, 250 )
 script = ExtResource( 1 )
-__meta__ = {
-"_edit_use_anchors_": false
-}
```
Poobslag, line 12: Revert this.

project/src/main/world/Cutscene.tscn
Poobslag: Reverted.

project/src/test/Tests.tscn
Poobslag: Reverted.

PR #1325, opened 2022-03-02
ChatTree only accepts one customer, chef.

Poobslag:
ChatTree accepted one chef id, but multiple customer ids before -- but,
only one customer id was actually used. It now only stores one customer
id.

Chatscript parser now warns if a chatscript file declares more than one
chef or customer.


project/src/main/ui/chat/career-cutscene-library.gd
```diff
@@ -96,9 +96,9 @@ func set_career_cutscene_root_path(new_career_cutscene_root_path: String) -> voi
 ##
```
Poobslag, line 96: Fix `customer_ids` parameter documentation.

Poobslag: Fixed, `The id of a creature who...`

PR #1360, opened 2022-03-31
Fixed patterny holes in flat unfrosted terrain

Poobslag:
Areas of flat unfrosted terrain only had two possible 4-way connected
tiles, and both had a hole in the middle. There are now six possible
tiles, and 50% of them will always be blank.


Poobslag:
The 'lemon block' tiles currently have two blank ones. Fix them to match the 'marsh block' tiles.

project/src/main/world/environment/frosted-overworld-tiler.gd
```diff
@@ -84,7 +84,10 @@ const BLOCK_AUTOTILE_COORDS_BY_BINDING := {
 	CAKE_RIGHT | CAKE_LEFT | CAKE_BOTTOM: [TileTypes.BLOCK,
 			[Vector2(4, 4), Vector2(5, 4)]],
 	CAKE_RIGHT | CAKE_LEFT | CAKE_BOTTOM | CAKE_TOP: [TileTypes.BLOCK,
-			[Vector2(0, 5), Vector2(1, 5)]],
+			# This first 4-way unfrosted tile is blank. We repeat it multiple times to increase its likelihood of
```
Poobslag, line 87: redundant 'repeat it multiple times'. if something is repeated, it's implicitly repeated multiple times

Poobslag: Removed 'multiple times'

Issue #94, opened 2020-03-31
Apply overworld enhancements to puzzle's restaurant scene


Poobslag:
The customer seating area should be rendered with the same engine as the overworld. Hopefully this should simplify some things as far as layers and stuff. It should make #51 and #54 more fixable too, since panning a camera around an environment is easier than scaling a set of sprites.

Make sure you don't accidentally break spatial sound effects.

Poobslag:
Ideally this would be absurdly seamless -- the player would run into the restaurant, and you'd just see sort of a splitscreen view, and the puzzle would overlay on top of it. It shouldn't even really feel like the player leaves the overworld.

Poobslag:
We've created a restaurant interior scene; we should reuse it for the puzzles

Poobslag:
Closed by #1376.

PR #1376, opened 2022-04-14
Puzzle now loads an environment instead of a custom restaurant scene.

Poobslag:
Before, puzzles used their own RestaurantPuzzleScene which used their
own 2D assets instead of the tilemap-based environments. Now they use
the same environments that are used in free roam, cutscenes and career
mode. These environments need to define spawn points for the chef and
customers and they can be used during puzzles.

The puzzle camera logic has been slightly redone. Before, the game
assumed three customers were at hard-coded coordinates separated by 1,000 pixels
horizontally. The new logic zooms from one customer to the other
customer regardless of where they are positioned. There was some logic
for keeping fat characters in frame -- this is now handled by a
'headroom' property.


Poobslag:
Closes #94 .

PR #1378, opened 2022-04-15
Removed shadow merging viewport logic.

Poobslag:
When shadows overlap, they now result in a darker shadow (which is bad.)
The old approach for fixing this involved creating a viewport the size
of the screen, rendering all the shadows to a viewport texture, and
altering the viewport texture's alpha. However this approach does not
work when there are multiple viewports into the same scene, such as what
happens during puzzles.


project/src/main/editor/creature/creature-shadows.gd
```diff
@@ -6,6 +6,7 @@ export (PackedScene) var CreatureShadowScene: PackedScene
 
 func _ready() -> void:
 	for creature in get_tree().get_nodes_in_group("creatures"):
+		print("9: create initial creature shadow %s" % [creature])
```
Poobslag, line 9: Remove (and run the delint script).

project/src/main/world/restaurant/RestaurantView.tscn
```diff
@@ -293,7 +293,7 @@ custom_styles/panel = SubResource( 4 )
 one_shot = true
 script = ExtResource( 12 )
 
-[connection signal="timeout" from="HelloTimer" to="HelloTimer" method="_on_timeout"]
 [connection signal="current_creature_index_changed" from="RestaurantViewport/Scene" to="." method="_on_RestaurantPuzzleScene_current_creature_index_changed"]
 [connection signal="current_creature_index_changed" from="RestaurantViewport/Scene" to="CustomerCameraMover" method="_on_RestaurantPuzzleScene_current_creature_index_changed"]
 [connection signal="food_eaten" from="RestaurantViewport/Scene" to="CustomerCameraMover" method="_on_RestaurantPuzzleScene_food_eaten"]
+[connection signal="timeout" from="HelloTimer" to="HelloTimer" method="_on_timeout"]
```
Poobslag, line 299: Revert, and run the delint script

project/src/main/editor/creature/creature-shadows.gd
Poobslag: Removed, and ran delint script

project/src/main/world/restaurant/RestaurantView.tscn
Poobslag: Reverted.

PR #1381, opened 2022-04-18
Wood background scaling fixes, renames.

Poobslag:
Wood background now stretches to the edges of the screen.

Loading screen and puzzle reuse the same scene for the wood background.

Renamed 'wood-backdrop' to less verbose 'wood-bg'. This is also
consistent with the 'Bg' terminology used elsewhere, like 'BgStrobe' and
'BgGlobViewports' in the puzzle scene. Backdrop is used throughout the
game for flat-colored panels used to fade things out.

Fixed PuzzleDemo's 'spawn pickups' key.


project/src/main/puzzle/puzzle.gd
```diff
@@ -30,7 +31,7 @@ func _ready() -> void:
 	get_customer().play_hello_voice()
 	
 	if CurrentLevel.settings.other.skip_intro:
-		$PuzzleMusicManager.start_puzzle_music()
+		$Fg/Fg/PuzzleMusicManager.start_puzzle_music()
```
Poobslag, line 34: $Fg/Fg?

Poobslag: Duplicated the bug this caused, and fixed it.

PR #1394, opened 2022-04-26
Change 'Regret...' condition for tutorials

Poobslag:
Previously, the 'Regret...' message showed up if we made the player
repeat a scenario. But there are times when we'll advance tutorials even
if they fail. The new 'Regret...' message logic keys off of
PuzzleState.level_performance.lost.

### DIFF
--- a/project/src/main/puzzle/tutorial/tutorial-squish-module.gd
+++ b/project/src/main/puzzle/tutorial/tutorial-squish-module.gd
@@ -16,14 +16,11 @@ var _show_diagram_count := 0
 
 ## set of level IDs which the player has attempted during this tutorial
 ## key: level id
-## value: true
-var _prepared_levels: Dictionary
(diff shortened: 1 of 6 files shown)


project/src/main/puzzle/tutorial/tutorial-squish-module.gd
```diff
@@ -17,14 +17,11 @@ var _show_diagram_count := 0
 ## set of level IDs which the player has attempted during this tutorial
 ## key: level id
 ## value: true
-var _prepared_levels: Dictionary
+var _level_attempt_count: Dictionary
```
Poobslag, line 20: Update this comment.

Poobslag: Updated to `number of attempts, initialized to '1' during the player's first attempt`

PR #1396, opened 2022-04-26
Fixed Tutorial's flash effect to fill the entire screen.

Poobslag:
This involved pulling the TutorialHud's flash effect into Puzzle's Canvas
node where it could fill the entire canvas. Instead of TutorialHud
having a HudFlash node, it now has a reference to one.

### DIFF
--- a/project/src/main/puzzle/Puzzle.tscn
+++ b/project/src/main/puzzle/Puzzle.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=99 format=2]
+[gd_scene load_steps=100 format=2]
 
 [ext_resource path="res://src/main/ui/menu/theme/h3.theme" type="Theme" id=1]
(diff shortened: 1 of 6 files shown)


project/src/main/puzzle/Puzzle.tscn
```diff
@@ -645,6 +647,9 @@ margin_top = 250.0
 margin_right = 482.0
 margin_bottom = 270.0
 
+[node name="HudFlash" parent="Hud" instance=ExtResource( 75 )]
+mouse_filter = 2
```
Poobslag, line 651: Push this mouse_filter down into the HudFlash scene.

Poobslag: Moved `mouse_filter = 2` into HudFlash.tscn.

PR #1410, opened 2022-05-06
LemonTree, FrameOverworldObstacle QOL improvements

Poobslag:
Lemon trees persist their horizontal flipped state.

Before, lemon trees were flipping themselves horizontally but this
information was stored in their child nodes in a transient way, and not
persisted to the scene.

Lemon trees now use odd/even numbers to represent whether the underlying
sprite should be flipped or not. This information is persisted to the
scenes. I've also reshuffled all of the lemon trees.

Added a range to LemonTree's leaf_type and mouth_type

Changing LemonTree's leaf/mouth properties now immediately updates its
appearance in the editor.

FrameOverworldObstacle now has a shuffle button.

FrameOverworldObstacle and LemonTree both use flip_h instead of scale.

### DIFF
--- a/project/src/main/world/environment/frame-overworld-obstacle.gd
+++ b/project/src/main/world/environment/frame-overworld-obstacle.gd
@@ -5,6 +5,29 @@ extends OverworldObstacle
 ## Current frame to display from the sprite sheet.
 export (int) var frame: int setget set_frame
 
+## If true, the sprite's texture is flipped horizontally.
+export (bool) var flip_h: bool setget set_flip_h
+
+## An editor toggle which randomizes the obstacle's appearance
+export (bool) var shuffle: bool setget set_shuffle
+
 func set_frame(new_frame: int) -> void:
 	frame = new_frame
 	$Sprite.frame = frame
+
+
+func set_flip_h(new_flip_h: bool) -> void:
+	flip_h = new_flip_h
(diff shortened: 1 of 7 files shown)


project/src/main/world/environment/frame-overworld-obstacle.gd
```diff
@@ -5,6 +5,28 @@ extends OverworldObstacle
 ## Current frame to display from the sprite sheet.
 export (int) var frame: int setget set_frame
 
+## If true, the sprite's texture is flipped horizontally.
+export (bool) var flip_h: bool setget set_flip_h
+
+export (bool) var shuffle: bool setget set_shuffle
```
Poobslag, line 11: Add a comment

Poobslag: Commented, `An editor toggle which randomizes the obstacle's appearance`

project/src/main/world/environment/lemon/LemonTree.tscn
```diff
@@ -305,24 +305,28 @@ __meta__ = {
 "_edit_group_": true,
 "shadow_scale": 1.0
 }
+leaf_type = 5
+mouth_type = 1
```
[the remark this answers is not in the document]
Poobslag: Reset.

PR #1414, opened 2022-05-09
Added/fixed 'RESET' animations

Poobslag:
Godot added RESET animations recently, but most of our animation players
did not utilize them correctly. AnimationPlayers now define a 'RESET'
animation which resets their properties in the editor. It is still not
used at runtime.


project/src/main/ui/chat/ChatChoiceButton.tscn
```diff
@@ -195,12 +258,11 @@ __meta__ = {
 fonts = [ ExtResource( 5 ), ExtResource( 9 ), ExtResource( 8 ), ExtResource( 3 ), ExtResource( 10 ) ]
 
 [node name="MoodSprite" type="Control" parent="."]
-modulate = Color( 1, 1, 1, 0.894118 )
 margin_left = -11.0
 margin_top = 13.0
 margin_right = 17.0
 margin_bottom = 41.0
-rect_scale = Vector2( 1.5, 1.5 )
+rect_rotation = 6.0
```
Poobslag, line 265: Reset this property.

Poobslag: Removed rect_rotation property and added it to the `RESET` animation.

PR #1432, opened 2022-05-17
Conditionally exclude cutscenes with sensei/restaurant

Poobslag:
Added logic to conditionally exclude cutscenes with Fat Sensei and the
restaurant. Early chapters (and perhaps later chapters) will have the
player by themselves, so we don't want a random cutscene which doesn't
make sense where Fat Sensei shows up.

### DIFF
--- a/project/src/main/ui/chat/chat-tree.gd
+++ b/project/src/main/ui/chat/chat-tree.gd
@@ -175,6 +175,20 @@ func reset() -> void:
 	_position.reset()
 
 
+## Returns 'true' if this cutscene is set inside the restaurant.
+##
+## This signifies that this cutscene should not be played if the player does not have a restaurant.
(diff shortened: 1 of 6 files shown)


project/src/main/ui/chat/chat-tree.gd
```diff
@@ -175,6 +175,14 @@ func reset() -> void:
 	_position.reset()
 
 
+func inside_restaurant() -> bool:
+	return location_id in ["marsh/inside_turbo_fat"]
+
+
+func has_sensei() -> bool:
+	return spawn_locations.has(CreatureLibrary.SENSEI_ID)
```
Poobslag, line 183: Add comments.

Poobslag: Added comments, `Returns true if...`

PR #1455, opened 2022-06-13
Music, money popups stay offscreen for tall resolutions

Poobslag:
Music and money popups stay offscreen. They were previously nested in a
fixed-size control with the rest of the UI elements. Those fixed-size UI
elements are now separated into a control called 'Center'.


project/src/main/puzzle/puzzle-money-hud.gd
```diff
@@ -0,0 +1,48 @@
+extends Control
+## Tween which shows/hides the money label which appears after a puzzle.
```
Poobslag, line 2: This comment is inaccurate.

Poobslag: Fixed comment, `Shows/hides the money label after a puzzle.`.

Also added comments to the rest of the script.

PR #1463, opened 2022-06-23
New chapter select screen for career mode.

Poobslag:
The chapter select screen replaces the functionality where the player
could use the arrow keys at the top of the screen to play earlier
chapters. When starting a new career session, they now explicitly choose
a chapter to play.

The chapter select screen supports paging, additional chapters will be
shown on additional pages which can be accessed with arrow buttons.
Chapters which haven't been reached yet appear in sepia tone. There are
chapter button graphics for Lemony Thickets and Merrymellow Marsh.
Additional graphics will be drawn as more chapters are added.

The chapter select screen shows a summary of the player's performance on
each chapter. Chapters are considered 'completed' when all their levels
have been beaten and all their cutscenes have been viewed. Before they
are completed, the summary will say something like 'You're 85% complete,
clear all the levels to reach 100%!'. After they are completed, the
summary will say something like 'You've gotten an S grade, and you're
88% of the way to an S+ grade!'

When replaying older chapters, the player will remain in that chapter
until it's beaten. This means when playing Lemony Thickets for example,
they won't progress past mile marker 9. Additionally, instead of only
two cutscenes for the 3rd and 6th level played, there is now an
additional cutscene for the 8th level played. This serves two purposes:
it lets them proceed through the story slightly faster, and provides an
incentive for playing the last two levels.


project/src/main/puzzle/rank-calculator.gd
```diff
@@ -477,6 +477,20 @@ static func grade(rank: float) -> String:
 	return grade
 
 
+## Calculates the next grade better than the specified grade.
+##
+## RankCalculator.next_grade("A-") = "A"
+## RankCalculator.next_grade("SSS") = "M"
+## RankCalculator.next_grade("M") = "M"
+## RankCalculator.next_grade("-") = "B-"
```
Poobslag, line 485: For consistency, this comment should omit the class name

Poobslag: And, the equals signs should be aligned

project/src/main/ui/career/career-distance.gd
```diff
@@ -1,55 +0,0 @@
...
-	_up_button.disabled = false
-	
-	if PlayerData.career.hours_passed > 0:
-		# disable the buttons; the player is in the middle of a career session
-		_down_button.disabled = true
-		_up_button.disabled = true
-	
-	var curr_region: CareerRegion = PlayerData.career.current_region()
-	var next_region: CareerRegion = PlayerData.career.next_region()
-	var prev_region: CareerRegion = PlayerData.career.prev_region()
```
Poobslag, line 28: The 'next_region' and 'prev_region' methods are unused, and should be removed

Poobslag: Removed `next_region()` and `prev_region()`. Verified there are no other references.

PR #1487, opened 2022-07-08
Reorder signal methods.

Poobslag:
'_on_foo()' functions should always be last. A few scripts had them out
of order.


Poobslag:
These were detected in IntelliJ with the regex `^func _on[\s\S]*^func _[^o]`. I tried to add them to the delint script but couldn't figure out how to make multiline regex expressions work with gnu grep.

PR #1494, opened 2022-07-11
Changed 'chat_theme_def' to 'chat_theme' globally. 375c.

Poobslag:
Before, we were storing dictionaries and only converting them to chat
themes on demand, resulting in effectively two distinct data models we
had to toggle between, and opening the door for mistyped strings instead
of strongly typed variable names.

We now store chat themes as objects throughout our code, they are only
stored as dictionaries briefly while being parsed or stored as json.

Changed 'chat_theme_def' in our json files to 'chat_theme' as well.
Upgraded all creatures to version 375c. This also requires a player dat
upgrade, as each save file defines the player's appearance and chat
theme.

### DIFF
--- a/project/src/main/ui/chat/chat-frame.gd
+++ b/project/src/main/ui/chat/chat-frame.gd
@@ -3,7 +3,7 @@ extends Control
 ## Window which displays a chat line.
 ##
 ## The chat window is decorated with objects in the background called 'accents'. These accents can be injected with the
-## chat_event.chat_theme_def property to configure the chat window's appearance.
+## chat_event.chat_theme property to configure the chat window's appearance.
 
 signal pop_out_completed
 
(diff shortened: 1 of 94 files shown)


project/src/main/world/creature/creature.gd
```diff
@@ -364,7 +364,7 @@ func get_creature_def() -> CreatureDef:
 	var result := CreatureDef.new()
 	result.creature_id = creature_id
 	result.dna = DnaUtils.trim_dna(dna)
-	result.chat_theme_def = chat_theme_def
+	result.chat_theme.from_json_dict(chat_theme.to_json_dict())
```
Poobslag, line 367: Add a comment describing why we don't pass by reference here

project/src/main/ui/chat/chat-frame.gd
```diff
@@ -59,7 +59,7 @@ func play_chat_event(chat_event: ChatEvent, squished: bool) -> void:
 			$SquishTween.unsquish()
 		_squished = squished
 	
-	var chat_theme := ChatTheme.new(chat_event.chat_theme_def)
+	var chat_theme := chat_event.chat_theme
```
Poobslag, line 62: We don't need this variable anymore

project/src/main/world/creature/creature.gd
Poobslag: Added comment, `# create a copy to prevent our chat theme from being modified accidentally`

PR #1501, opened 2022-07-14
Bug: successful_levels entries missing from finished_levels

Poobslag:
A typo in the LevelHistory script made it so that if the player was
successful in a level (e.g clearing 100 lines in Marathon Normal) the
level was added to LevelHistory.successful_levels, but not to
LevelHistory.finished_levels. I've fixed the bug.

This bug affects player save data, as any finished levels in earlier
versions were not recorded. I've upgraded the player data to 3776 and
added a translation step which fills any gaps in finished_levels with
their entries from successful_levels.

### DIFF
--- a/project/src/main/save-item-upgrader.gd
+++ b/project/src/main/save-item-upgrader.gd
@@ -31,9 +31,9 @@ var current_version := ""
 ## SaveItemUpgrader does not have logic for upgrading specific save data versions. This upgrade logic must be defined
 ## on an external object and incorporated via this 'add_upgrade_method' method.
 ##
-## The specified upgrade method should accept a SaveData object and return a modified SaveData object. The upgrade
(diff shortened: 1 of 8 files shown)


project/src/main/save-item-upgrader.gd
```diff
@@ -87,7 +87,7 @@ func upgrade(json_save_items: Array) -> Array:
 				"version":
 					save_item.value = upgrade_method.new_version
 				_:
-					save_item = upgrade_method.object.call(upgrade_method.method, save_item)
+					save_item = upgrade_method.object.call(upgrade_method.method, old_save_items, save_item)
```
Poobslag, line 90: Update this method comment to describe this new parameter.

Poobslag: Updated, `should accept an Array of source data and a SaveData object...`

PR #1514, opened 2022-07-18
Practice menu redesign with career levels.

Poobslag:
Redesigned the practice menu. The new practice menu lets the player pick
both training modes and career levels from the same screen.

Different difficulties of training modes (e.g. Sprint Normal, Sprint Expert) are
now treated as different levels, instead of harder difficulties of the
same level.

The difficulty slider has been replaced with a speed slider.
This means the player can, for example, practice Ultra Expert at FFF
speed instead of F0 speed, which will give them a huge speed advantage
if they can keep up with it. This arguably ruins the level playing field
where all players were forced to play on the same speed, but it seems
analogous to NES Tetris allowing players to pick a starting level of 18
to achieve an unfairly high score compared to someone who started at
level 0.

Training levels are now placed in their own regions, there is a 'Sprint
Training Level' region for example. These are all categorized as
'OtherRegions' instead of 'CareerRegions' as they are lacking properties
such as distance. They are managed by a new 'OtherRegionsLibrary'.

The new practice menu remembers the player's previous level choice, this
is stored in their save data.


project/assets/test/ui/level-select/other-regions-simple.json
```diff
@@ -0,0 +1,23 @@
+{
+  "regions": [
```
Poobslag, line 2: These regions should have an 'id'.

project/src/main/ui/menu/practice-menu.gd
```diff
@@ -1,147 +1,199 @@
 extends Control
-## Scene which lets the player repeatedly play a set of levels.
```
Poobslag, line 2: Add top-level comment.

project/src/main/ui/menu/practice-menu.gd
```diff
@@ -1,147 +1,199 @@
...
+func _assign_default_recent_data() -> void:
+	if PlayerData.practice.region_id and PlayerData.practice.level_id:
+		# player has already launched the practice menu
+		return
+	
+	PlayerData.practice.region_id = ""
+	PlayerData.practice.level_id = ""
+	
+	PlayerData.practice.region_id = OtherRegion.ID_MARATHON
+	PlayerData.practice.level_id = "practice/marathon_normal"
```
Poobslag, line 63: This code is very strange, why do you assign these fields twice?

project/src/main/ui/menu/practice-menu.gd
```diff
@@ -1,147 +1,199 @@
...
+func _refresh_speed_selector() -> void:
+	if _region is CareerRegion:
+		# populate speed selector with region speeds
+		var min_index := PieceSpeeds.speed_ids.find(_region.min_piece_speed)
+		var max_index := PieceSpeeds.speed_ids.find(_region.max_piece_speed)
+		var new_speed_names := PieceSpeeds.speed_ids.slice(min_index, max_index)
+		_speed_selector.set_speed_names(new_speed_names)
+		_speed_selector.set_editable(true)
+	else:
+		# populate speed selector with arbitrary speeds
```
Poobslag, line 127: rephrase `arbitrary speeds` -- maybe something like `a range of speeds specific to that level`

project/assets/test/ui/level-select/other-regions-simple.json
Poobslag: Added `id` fields

project/src/demo/puzzle/PuzzleDemo.tscn
```diff
@@ -5,6 +5,6 @@
 
 [node name="PuzzleDemo" type="Node"]
 script = ExtResource( 2 )
-level_path = "res://assets/main/puzzle/levels/practice/marathon-normal.json"
+level_path = "res://assets/main/puzzle/levels/career/paranormal-tetrominon.json"
```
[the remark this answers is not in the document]
Poobslag: Reverted

project/src/main/ui/menu/paged-level-buttons.gd
```diff
@@ -0,0 +1,225 @@
+class_name PagedLevelButtons
+extends HBoxContainer
+## Creates and arranges level buttons for the level select screen.
+##
+## These buttons are arranged in multiple pages which can be navigated with arrow buttons.
+
+const LEVELS_PER_PAGE := 18
```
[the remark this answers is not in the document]
Poobslag: Renamed

project/src/main/ui/menu/practice-menu.gd
Poobslag: Added comment, `## Scene which lets the player repeatedly play a set of levels.`

project/src/main/ui/menu/practice-menu.gd
```diff
@@ -1,147 +1,202 @@
...
+			# hard training levels by slowing them down
+			speed_matrix_row = speed_matrix_row.slice(speed_matrix_row.find(start_speed), speed_matrix_row.size() - 1)
+		
+		if _region.id == OtherRegion.ID_RANK:
+			# Players can't change the speed in rank mode.
+			speed_editable = false
+		
+		_speed_selector.set_speed_names(speed_matrix_row)
+		_speed_selector.set_selected_speed(start_speed)
+		_speed_selector.set_editable(speed_editable)
```
Poobslag, line 149: This code could be more symmetrical, why not pull this piece out of the `if/else` block

Poobslag: Refactored code, pulling the variables and assignments outside of the if block

PR #1518, opened 2022-07-19
Tutorial now uses paged panel instead of scrollable panel.

Poobslag:
The scrollable panel is going away soon, it is only used in FreeRoam
mode. There are also not enough tutorials for it to make sense.

### DIFF
--- a/project/assets/main/puzzle/other-regions.json
+++ b/project/assets/main/puzzle/other-regions.json
@@ -4,9 +4,6 @@
       "id": "tutorial",
       "name": "Tutorials",
       "description": "Tutorials designed to teach new concepts to players of all skill levels.",
-      "flags": [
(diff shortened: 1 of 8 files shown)


project/assets/main/puzzle/other-regions.json
```diff
@@ -4,9 +4,6 @@
       "id": "tutorial",
       "name": "Tutorials",
       "description": "Tutorials designed to teach new concepts to players of all skill levels.",
-      "flags": [
-        "tutorial"
-      ],
```
Poobslag, line 9: Mention this in the commit comment.

Poobslag: Added comment, `Replaced tutorial flag with TUTORIAL_ID constant`

PR #1527, opened 2022-07-25
Higher resolution combo accent texture.

Poobslag:
The old texture looked a little jagged and pixelly. We are reusing this
texture for the upcoming spin move notifications at a larger size, so it
is even more noticable when it is larger.


Poobslag:
Old:

image

New:

image

The difference is marginal at this size, but it will become more noticable when the textures are utilized at a larger size in the upcoming spin move update.

PR #1560, opened 2022-08-16
Chatscript can set boolean/string flags.

Poobslag:
These flags are a way to store conversation paths the player took.
Boolean flags are suitable for 'did the player do this thing or not?'
String flags are suitable for 'which choice did the player make?'

Flags can be manipulated with the following chatscript meta items:

    set_flag foo ----- set a flag 'foo' to 'true'
    set_flag foo bar - set a flag 'foo' to 'bar'
    unset_flag foo --- unset a flag 'foo'

Flags can be evaluated with the following chatscript boolean tokens:

    is_flag foo ------ true if the flag 'foo' is truthy
    is_flag foo bar -- true if the flag 'foo' holds a value 'bar'


project/src/main/ui/chat/chat-tree.gd
```diff
@@ -229,3 +231,26 @@ func _apply_say_if_conditions() -> void:
 			# advance through the current chat branch
 			_position.index += 1
 			did_increment = true
+
+
+## Set any flags corresponding to 'set_flag' statements.
```
Poobslag, line 236: Also applies `unset_flag` statements, should mention it in the comment

Poobslag: Expanded comment, `Unsets flags for 'unset_flag' statements.`

PR #1563, opened 2022-08-16
Chatscript supports phrases.

Poobslag:
Phrases can be set with 'set_phrase foo Bar'. The phrase values can have
spaces. These are used for storing the player's restaurant name
selection.


project/src/main/ui/chat/chat-history.gd
```diff
@@ -11,19 +11,30 @@ const CHAT_AGE_NEVER := 99999999
...
+## key: (String) phrase name
+## value: (String) phrase value
+var phrases: Dictionary
+
+## Flags which can be stored/retreived from chatscript. Flags represent chat choices the player made, such as the
+## answer to a yes or no question. They are persisted alongside the player's chat history.
+##
+## key: (String) flag name
+## value: (String) flag value
+var flags: Dictionary
```
Poobslag, line 30: Move this up to preserve alphabetical order. (I think you reordered these when some of the fields were private.)

Poobslag: Moved.

PR #1576, opened 2022-08-24
Update two-digit chat history entries

Poobslag:
Two-digit cutscenes were expanded to three digits in fd6b909f, but the
player's chat history was not updated. This meant players who'd played
the game previously would lose all their cutscene progress and have to
rewatch all the same cutscenes again.

### DIFF
--- a/project/src/main/player-save-upgrader.gd
+++ b/project/src/main/player-save-upgrader.gd
@@ -240,10 +240,15 @@ const PREFIX_REPLACEMENTS_2743 := {
 }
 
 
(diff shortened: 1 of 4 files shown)


project/src/main/player-save-upgrader.gd
```diff
@@ -240,10 +240,14 @@ const PREFIX_REPLACEMENTS_2743 := {
 }
 
 
+var _third_zero_regex: RegEx
```
Poobslag, line 243: Add code comment

Poobslag: Added comment, `Regex used for replacing two ...`

PR #1589, opened 2022-08-29
Fixed bug where intro level cutscenes could be skipped

Poobslag:
Intro level cutscenes were dependent on whether the player had played
that specific level before. This meant intro cinematics could be skipped
if the intro level was recycled or if the player tried it in practice
mode.

The intro level cutscenes are now dependent on whether the player has progressed
through career mode or not. This makes many of the existing career tests
redundant or overly convoluted, such as tests where the player progresses
through several boss levels in regions with no intro level. I've removed these
tests, because I anticipate most regions having an intro level anyways so the
tests aren't worth preserving.

### DIFF
--- a/project/src/main/career/career-calendar.gd
+++ b/project/src/main/career/career-calendar.gd
@@ -25,13 +25,21 @@ func advance_clock(new_distance_earned: int, success: bool) -> void:
 	career_data.hours_passed += 1
 	
 	if career_data.is_boss_level():
-		var boss_region: CareerRegion = career_data.current_region()
+		var region: CareerRegion = career_data.current_region()
 		if success:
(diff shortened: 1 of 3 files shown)


project/src/main/career/career-calendar.gd
```diff
@@ -61,7 +69,7 @@ func _apply_distance_earned(unapplied_distance_earned: int) -> int:
 	var newly_travelled_distance := unapplied_distance_earned
 	
 	if not career_data.is_region_cleared(region) and region.intro_level \
-			and not career_data.is_intro_level_finished(region):
+			and career_data.best_distance_travelled <= region.start:
```
Poobslag, line 72: Use your new utility method here.

PR #1599, opened 2022-09-01
ResourceCache: threaded, load_seconds, low_priority

Poobslag:
Added ResourceCache.threaded flag. ResourceCache can be configured to be
threaded or non-threaded. This allows us to disable threading even for
targets which support it.

Added ResourceCache.load_seconds flag. ResourceCache can be configured
to load resources gradually. This makes the loading screen less
stuttery.

Added ResourceCache.low_priority_resource_paths parameter. ResourceCache
has a few resources such as 'Puzzle.tscn' which, if loaded too early,
take a long time to cache because they load other scenes and resources.
Loading them last results in fewer stutters when caching resources.

### DIFF
--- a/project/src/main/resource-cache.gd
+++ b/project/src/main/resource-cache.gd
@@ -16,8 +16,8 @@ const THREAD_COUNT := 4
 ## loading scenes is slower than loading regular resources; this constant estimates how much slower
 const WORK_PER_SCENE := 8.0
 
-## in non-threaded environments, this is how long we're willing to block a thread to cache resources
-const CHUNK_SECONDS := 0.1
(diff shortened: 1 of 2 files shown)


project/src/main/resource-cache.gd
```diff
@@ -256,6 +307,15 @@ func _find_resource_paths() -> Array:
 	# We shuffle the pngs to prevent clumps of similar files. We use a known seed to keep the timing predictable.
 	_remaining_resource_paths.shuffle()
 	_remaining_scene_paths.shuffle()
+	
+	for low_priority_resource_path in low_priority_resource_paths:
```
Poobslag, line 311: Add a comment

PR #1618, opened 2022-09-06
Use 'obstructed' instead of 'blocked'.

Poobslag:
'Block' is almost always used to refer to puzzle playfield blocks, so
using the word 'blocked' results in confusing sentences like 'if the
target block is blocked'

'Obstruct' is less ambiguous.


project/assets/main/locale/es.po
```diff
@@ -272,7 +272,7 @@ msgid "Cookies"
 msgstr ""
 
 #: project/assets/main/locale/localizables-extracted.py:60
-msgid "Let's make some cookies! Who brought the U pieces?"
+msgid "Let's make some cookies! Who brought the U-Blocks?"
```
Poobslag, line 275: Revert this.

Poobslag: Reverted.

PR #1641, opened 2022-09-16
Added 'customer_if' flags

Poobslag:
Creatures can now define a customer_if flag. It defaults to 'true' so all existing creature definitions are customers. Setting it to 'false' will prevent them from appearing as a random customer. It can also be set to an expression like 'region_cleared marsh' which will make the customer appear once the criteria are met. (Or, prevent the customer from appearing once the criteria stop being met.)

### DIFF
--- a/project/assets/main/creatures/primary/bones.json
+++ b/project/assets/main/creatures/primary/bones.json
@@ -34,5 +34,6 @@
     "accent_texture": 2,
     "color": "1a1a1e",
     "dark": true
-  }
+  },
+  "customer_if": "false"
 }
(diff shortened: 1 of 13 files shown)


project/src/main/editor/creature/CreatureEditor.tscn
```diff
@@ -1983,9 +1980,6 @@ margin_bottom = 480.0
 popup_exclusive = true
 window_title = "Export Creature"
 access = 2
-current_dir = "/"
-current_file = "509e7c82-9399-425a-9f15-9370c2b3de8b"
-current_path = "/509e7c82-9399-425a-9f15-9370c2b3de8b"
```
Poobslag, line 1988: Revert.

Poobslag: Reverted.

PR #1663, opened 2022-09-27
Extracted ConfigStringUtils.

Poobslag:
This breaks the AfterLineCleared phase condition into three simple easy-to-understand parts -- translating strings to int arrays, inverting the int arrays, and combining those two pieces in the AfterLineCleared phase condition. The first two parts are generic and can be reused by other phase conditions or effects in the future.

### DIFF
--- a/project/src/main/puzzle/level/config-string-utils.gd
+++ b/project/src/main/puzzle/level/config-string-utils.gd
@@ -1,0 +1,98 @@
+class_name ConfigStringUtils
+## Utilities for level config strings.
+
+## Converts a series of ints like [2, 3, 4, 5, 7] into a config string like '2-5,7'.
+##
+## config_string_from_ints([2, 3, 4])       = '2-4'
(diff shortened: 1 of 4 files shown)


project/src/main/puzzle/level/config-string-utils.gd
```diff
@@ -0,0 +1,98 @@
...
+			hi = int(sub_expression) + 1
+		for y in range(lo, hi):
+			ints[y] = true
+	
+	var result := ints.keys()
+	result.sort()
+	return result
+
+
+## Converts an array of user-friendly line indexes like [0, 1] into row numbers like [19, 18] or vice-versa.
```
Poobslag, line 82: Instead of 'line indexes', specify 'puzzle row indexes'

Lines can refer to a lot of things, like lines of code.

Poobslag: Changed comment, and also renamed the method

PR #1664, opened 2022-09-27
Added SfxDeconflicter to avoid playing two sounds simultaneously.

### LINKED ISSUE #990: Cutscenes should avoid playing too many sound effects at once
Some scenes like [goodbye_everyone_100.chat](https://github.com/Poobslag/turbofat/blob/9269f6712dcf35aadb3c5d80f8a50a3c769c7ea7/project/assets/main/puzzle/levels/cutscenes/marsh/goodbye-everyone-100.chat) have moments where everyone waves, or everyone smiles. This results in a lot of simultaneous sound effects which is unpleasant.

Cutscenes should detect overlapping sound effects, and only play one (the character speaking.)

Poobslag:
This affects cutscenes where multiple creatures might emote the same face at the same time. Instead of playing a very loud version of the emote sound, SfxDeconflicter only plays one instance of the sound so the player hears it at a normal volume.

This node is generic and can apply to any node which uses an AudioStreamPlayer to play a sound effect.

Fixes #990.


project/project.godot
```diff
@@ -1355,9 +1355,11 @@ PlayerSave="*res://src/main/data/player-save.gd"
 PuzzleState="*res://src/main/puzzle/puzzle-state.gd"
 ResourceCache="*res://src/main/utils/ResourceCache.tscn"
 SceneTransition="*res://src/main/utils/scene-transition.gd"
+SfxDeconflicter="*res://src/main/utils/SfxDeconflicter.tscn"
 SystemData="*res://src/main/data/system-data.gd"
 SystemSave="*res://src/main/data/system-save.gd"
 PhaseConditions="*res://src/main/puzzle/level/phase-conditions.gd"
+AudioStreamDeconflicter="*res://src/main/utils/SfxDeconflicter.tscn"
```
Poobslag, line 1362: Remove redundant 'AudioStreamDeconflicter'

PR #1668, opened 2022-09-27
Fixed bug where moles could spawn behind the active piece.

Poobslag:


project/src/main/puzzle/critter/moles.gd
```diff
@@ -164,6 +165,15 @@ func _mole_cell_has_mole(mole_cell: Vector2) -> bool:
 	return _moles_by_cell.has(mole_cell)
 
 
+func _mole_cell_overlaps_piece(mole_cell: Vector2, piece: ActivePiece) -> bool:
```
Poobslag, line 168: Add comment

PR #1670, opened 2022-09-28
Add unused career levels; perfect_fit, fruit_on_the_top

Poobslag:
These were added to the career folder, but never assigned regions. I've updated the ReleaseToolkit to detect unused levels.

### DIFF
--- a/project/src/main/career/career-level-library.gd
+++ b/project/src/main/career/career-level-library.gd
@@ -24,6 +24,10 @@ func all_level_ids() -> Array:
 	for region in regions:
 		for career_level in region.levels:
 			result[career_level.level_id] = true
+		if region.boss_level:
+			result[region.boss_level.level_id] = true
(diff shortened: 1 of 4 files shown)


project/src/main/career/career-level-library.gd
```diff
@@ -24,6 +24,10 @@ func all_level_ids() -> Array:
 	for region in regions:
 		for career_level in region.levels:
 			result[career_level.level_id] = true
+		if region.boss_level:
+			result[region.boss_level.level_id] = true
+		if region.intro_level:
+			result[region.intro_level.level_id] = true
```
Poobslag, line 30: Document this change.

Poobslag: Added commit comment, `CareerLevelLibrary.all_level_ids now includes boss, intro level`

PR #1672, opened 2022-09-28
Renamed level trigger phases: 'line_cleared', 'piece_written'

Poobslag:
These phases explicitly specified 'AFTER line cleared' and 'AFTER piece written' in anticipation of other more granular phase conditions being written some day. However, I don't see a need for those.

Additionally, the timing of these triggers needs to be very specific to avoid unwanted side effects. The 'after line cleared' phase condition used to coincide with the after_line_cleared signal, but they now occur at different times, so I'm renaming these phase conditions to avoid confusion.

### DIFF
--- a/project/src/main/puzzle/line-clearer.gd
+++ b/project/src/main/puzzle/line-clearer.gd
@@ -117,7 +117,7 @@ func clear_line(y: int, total_lines: int, remaining_lines: int) -> void:
 func schedule_full_row_line_clears() -> void:
 	var lines_to_clear := []
 	for y in range(PuzzleTileMap.ROW_COUNT):
-		if _tile_map.playfield_row_is_full(y):
+		if _tile_map.row_is_full(y):
 			lines_to_clear.append(y)
 	if lines_to_clear:
 		schedule_line_clears(lines_to_clear, PieceSpeeds.current_speed.line_clear_delay)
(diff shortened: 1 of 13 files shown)


project/src/main/puzzle/line-clearer.gd
```diff
@@ -213,7 +213,7 @@ func _delete_lines(old_lines_being_cleared: Array, _old_lines_being_erased: Arra
 	old_lines_being_deleted.sort()
 	var max_line: int = old_lines_being_deleted.back() if old_lines_being_deleted else -1
 	for y in range(max_line):
-		if not _tile_map.playfield_row_is_empty(y) and not old_lines_being_deleted.has(y):
+		if not _tile_map.row_is_empty(y) and not old_lines_being_deleted.has(y):
```
Poobslag, line 216: Document this rename too.

project/src/main/puzzle/puzzle-tile-map.gd
```diff
@@ -269,7 +269,7 @@ func _disconnect_block(pos: Vector2, dir_mask: int = 15) -> void:
 ## 	'bottom_row': The lowest row to shift. All rows at or above this row will be shifted.
 ##
 ## 	'direction': The direction to shift the rows, such as Vector2.UP or Vector2.DOWN.
-func _shift_rows(bottom_row: int, direction: Vector2) -> void:
+func shift_rows(bottom_row: int, direction: Vector2) -> void:
```
Poobslag, line 272: Document this.

project/src/main/puzzle/line-clearer.gd
Poobslag: Comment, `Renamed PuzzleTileMap.playfield_row_ methods. This is redundant.`

PR #1677, opened 2022-09-30
New 'unlock' cheat on practice menu

Poobslag:


project/src/main/ui/menu/paged-level-buttons.gd
```diff
@@ -32,6 +32,8 @@ var _page := 0
 ## value: (LevelSettings) level settings for the level id
 var _level_settings_by_id: Dictionary = {}
 
+var _unlock_cheat_enabled := false
```
Poobslag, line 35: comment

Poobslag: added comment `temporarily unlocked all levels with a cheat ..`

PR #1678, opened 2022-09-30
New level triggers: 'box built', 'line_cleared n=5'

Poobslag:
This 'box built' phase trigger lets us create levels where special things happen when the player makes snack boxes or cake boxes (or both.) In the future this could be expanded to reward specific boxes (like JLO boxes) but this feels too obscure to be fun.

This 'line cleared' phase trigger lets us create levels where surprises happen after clearing a certain number of lines. Or, after clearing 5, 10, 15... lines.

### DIFF
--- a/project/src/main/puzzle/level/phase-conditions.gd
+++ b/project/src/main/puzzle/level/phase-conditions.gd
@@ -7,6 +7,7 @@ class PieceWrittenPhaseCondition extends PhaseCondition:
 	## We precalculate which pieces will trigger the rule, up to this number of pieces.
 	## 20,000 corresponds to an expert player playing at ~200 PPM for two hours.
 	const MAX_PIECE_INDEX := 20000
+	const MAX_LINE_INDEX := 10000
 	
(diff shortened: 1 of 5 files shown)


project/src/main/puzzle/level/phase-conditions.gd
```diff
@@ -71,11 +72,57 @@ class PieceWrittenPhaseCondition extends PhaseCondition:
...
+
+
 class LineClearedPhaseCondition extends PhaseCondition:
+	
 	## key: (int) a line which causes the trigger to fire when cleared. 0 is the highest line in the playfield.
 	## value: (bool) true
 	var which_lines := {}
 	
+	var how_many_lines_string := ""
+	var how_many_lines := {}
```
Poobslag, line 124: Comments

Poobslag: Added comment, `key: (int) a line milestone which causes the trigger to fire when cleared.`

PR #1684, opened 2022-10-05
Fixed bug with mole pickups overlapping the active piece.

Poobslag:
Fixed a bug where if the mole dug up a pickup while a piece was overlapping them, the pickup would appear 'behind' the piece.

### DIFF
--- a/project/src/main/puzzle/pickups.gd
+++ b/project/src/main/puzzle/pickups.gd
@@ -14,7 +14,8 @@ const PICKUP_DEFAULT: int = BlocksDuringRules.PickupType.DEFAULT
 const PICKUP_FLOAT: int = BlocksDuringRules.PickupType.FLOAT
 const PICKUP_FLOAT_REGEN: int = BlocksDuringRules.PickupType.FLOAT_REGEN
 
-export (NodePath) var _puzzle_tile_map_path: NodePath
+export (NodePath) var piece_manager_path: NodePath setget set_piece_manager_path
+export (NodePath) var puzzle_tile_map_path: NodePath
 export (PackedScene) var PickupScene: PackedScene
 
 ## key: (Vector2) playfield cell positions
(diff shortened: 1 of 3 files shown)


project/src/main/puzzle/pickups.gd
```diff
@@ -89,6 +96,24 @@ func clear() -> void:
...
+	if not is_inside_tree():
+		return
+	
+	if _piece_manager:
+		_piece_manager.disconnect("piece_disturbed", self, "_on_PieceManager_piece_disturbed")
+	
+	_piece_manager = get_node(piece_manager_path) if piece_manager_path else null
+	
+	if _piece_manager:
+		_piece_manager.connect("piece_disturbed", self, "_on_PieceManager_piece_disturbed")
```
Poobslag, line 114: You are connecting a non-existent signal.

Poobslag: The signal receiver method already existed and was not shown in this diff, but it exists.

project/src/main/puzzle/pickups.gd
```diff
@@ -27,7 +28,10 @@ var _pickup_sfx_index := 0
 ## how many more pickup sounds should play after the current one
 var _remaining_pickup_sfx := 0
 
-onready var _puzzle_tile_map: PuzzleTileMap = get_node(_puzzle_tile_map_path)
+onready var _puzzle_tile_map: PuzzleTileMap = get_node(puzzle_tile_map_path)
```
Poobslag, line 31: This should not be assigned onready, because piece_manager_path can (and will) be null in the editor.

Poobslag: You're confusing PuzzleTileMap for PieceManagerPath; `puzzle_tile_map_path` is always assigned in the editor.

PR #1708, opened 2022-10-14
Replace top out rank penalty with money penalty

Poobslag:
Topping out no longer affects the rank algorithm. Instead, it takes away ¥100 and plays a long 'top out' animation which takes some time anyways. This is enough of a penalty.

The score panel now shows a red number and particle effect when the player tops out.

### DIFF
--- a/project/src/main/puzzle/score-value-label.gd
+++ b/project/src/main/puzzle/score-value-label.gd
@@ -1,10 +1,33 @@
 extends Label
 ## Displays the player's score.
 
+onready var _top_out_particles := $TopOutParticles
+
 func _ready() -> void:
 	PuzzleState.connect("score_changed", self, "_on_PuzzleState_score_changed")
+	PuzzleState.connect("topped_out", self, "_on_PuzzleState_topped_out")
 	text = StringUtils.format_money(0)
+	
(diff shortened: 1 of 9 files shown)


project/src/main/puzzle/Puzzle.tscn
```diff
@@ -144,7 +144,7 @@ initial_velocity_random = 0.5
 orbit_velocity = 0.0
 orbit_velocity_random = 0.0
 scale = 5.0
-color_ramp = SubResource( 4 )
+color_ramp = SubResource( 25 )
```
Poobslag, line 147: rv

project/src/main/puzzle/score-value-label.gd
```diff
@@ -1,10 +1,33 @@
 extends Label
 ## Displays the player's score.
 
+onready var _score_particles := $ScoreParticles
```
Poobslag, line 4: comments

project/src/main/puzzle/Puzzle.tscn
Poobslag: Reverted.

project/src/main/puzzle/score-value-label.gd
Poobslag: Changed field/node name to `_topout_particles` precluding the need for a comment.

PR #1739, opened 2022-10-29
Carrot improvements: sfx, deconflicting

Poobslag:
Added carrot sfx. They make a 'whoosh' sound when they poof in and out, and a cartoon plane sound when they fly by.

Fixed carrot deconflicting algorithm. This algorithm used to swap carrots which were too close together, but this often just moved other carrots which were also too close. The new algorithm is more consistent, as demonstrated with a unit test.

Fixed remove_carrots() performance degradation. When calling it with a large number like 999,999, it would still loop 999,999 times even if it only had to remove 3 carrots. It now terminates the loop prematurely once the last carrot is removed.

### DIFF
--- a/project/src/main/puzzle/critter/Carrot.tscn
+++ b/project/src/main/puzzle/critter/Carrot.tscn
@@ -21,6 +21,7 @@ colors = PoolColorArray( 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0 )
 gradient = SubResource( 6 )
 
 [sub_resource type="ParticlesMaterial" id=5]
+resource_local_to_scene = true
 emission_shape = 2
 emission_box_extents = Vector3( 10, 4, 1 )
 flag_disable_z = true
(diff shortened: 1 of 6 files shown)


project/src/main/puzzle/critter/carrot.gd
```diff
@@ -5,6 +5,8 @@ extends Node2D
 ## Carrots remain onscreen for several seconds. They have many different sizes, and can also leave behind a smokescreen
 ## which blocks the player's vision for even longer.
 
+signal started_hiding
```
Poobslag, line 8: Document this signal.

project/src/main/puzzle/critter/Carrot.tscn
```diff
@@ -21,6 +21,7 @@ colors = PoolColorArray( 1, 1, 1, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0 )
 gradient = SubResource( 6 )
 
 [sub_resource type="ParticlesMaterial" id=5]
+resource_local_to_scene = true
```
Poobslag, line 24: Document this change.

project/src/main/puzzle/critter/carrot.gd
Poobslag: Commented, ` ## Emitted when the carrot starts ...`

project/src/main/puzzle/critter/Carrot.tscn
Poobslag: Documented, `where the smoke size would increase/decrease..../`

PR #1764, opened 2022-11-08
Avoid reporting 'All Clear!' between tutorial sections.

Poobslag:
The 'make cakes' tutorial clears all cake lines at the end of a tutorial section. This would trigger the all clear message if the player hadn't placed their final piece.

I've added a 'tutorial_section_finished' flag which is assigned/unassigned during tutorials. If it's been set, the all clear message is suppressed.

### DIFF
--- a/project/src/main/puzzle/puzzle-state.gd
+++ b/project/src/main/puzzle/puzzle-state.gd
@@ -86,6 +86,9 @@ var fatness_score := 0
 ## 'true' if the player has started a game which is still running.
 var game_active: bool
 
+## 'true' if the player finished this tutorial section.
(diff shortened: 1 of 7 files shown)


project/src/main/puzzle/puzzle-state.gd
```diff
@@ -273,13 +277,15 @@ func end_combo() -> void:
 
 ## Reset all score data, such as when starting a level over.
 func reset() -> void:
+
```
Poobslag, line 280: rv

Poobslag: Reverted.

PR #1774, opened 2022-11-11
Fixed bugs when interrupting/restarting tutorials.

### LINKED ISSUE #1180: Tutorials behave strangely if you restart in the middle
If you restart in the middle of a tutorial at certain crucial times, the tutorial will behave strangely -- either restarting from the middle, soft locking, or letting you manipulate pieces while the menu is active.

If you clear a tutorial section and quickly restart, the tutorial will behave strangely, restarting from the middle of the tutorial.

If you fail a tutorial section and quickly retry, the game behaves strangely as well.

This is because we're using yield statements which don't reset their state when the tutorial resets. We shouldn't use yield statements, we should use something like timers.

Poobslag:
When restarting tutorials after the 'Good!' message showed up, the piece would never appear, softlocking the tutorial. This is because PieceManager set physics_process to false during the level transition but did not set it back to true. It now sets physics_process to true at the start of a new game.

Tutorials would demonstrate weird behavior when interrupted, because 'yield' statements would continue from the previously run tutorial. These 'yield' statements have been replaced with timers and listeners which are cleaned up when a new tutorial is started.

The choices which appear when the player is watching tutorial diagrams were lingering even if the player gave up on the tutorial. This let them do strange things like choose tutorial options when they weren't playing anymore. The choices are now hidden when the tutorial ends.

The TutorialModule framework now includes utility methods to make it easier for tutorials to do something after a delay, or do something after all messages are shown to the player.

Closes #1180.

### DIFF
--- a/project/src/main/puzzle/tutorial/tutorial-module.gd
+++ b/project/src/main/puzzle/tutorial/tutorial-module.gd
@@ -4,11 +4,19 @@ extends Node
 ##
 ## Subclasses can show messages and advances the player through the tutorial as they complete tasks.
 
+## if 'true', the next level change will be accompanied by a 'Ready? Go!' countdown
+var _start_customer_countdown := false
+
 ## generic nodes used by tutorial module subclasses
 var hud: TutorialHud
 var puzzle: Puzzle
 var playfield: Playfield
 var piece_manager: PieceManager
+
+## Holds all temporary timers. These timers are not created by get_tree().create_timer() because we need to clean them
+## up if the tutorial is interrupted. Otherwise for example, we might schedule a level change 3 seconds from now, but
+## then the player restarts the tutorial, and the level change happens after they've restarted.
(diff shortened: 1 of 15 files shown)


project/src/main/puzzle/tutorial/tutorial-module.gd
```diff
@@ -24,14 +34,33 @@ func _ready() -> void:
...
+
+
+func _on_PuzzleState_game_ended() -> void:
+	_start_customer_countdown = false
+	cleanup_listeners()
+
+
+func _on_PuzzleState_game_prepared() -> void:
+	_start_customer_countdown = false
+	cleanup_listeners()
```
Poobslag, line 56: Move listener methods to end.

Poobslag: Moved.

PR #1793, opened 2022-11-24
Localize chat messages with symbols, 'Hello #player#'

Poobslag:
Before, messages like 'I'm #mood#' would fail
to be translated for two reasons.

Firstly, the translation happened when the message was displayed to the user, after all entities were replaced. So even if we had a translation for 'I'm #mood#', the translation engine would instead search for 'I'm angry' which was undefined.

Secondly, the translation engine did not translate replaceable entities like 'angry'.

Both these issues are now fixed.

### DIFF
--- a/project/src/main/chat/chat-tree.gd
+++ b/project/src/main/chat/chat-tree.gd
@@ -246,7 +246,12 @@ func _assign_flags_and_phrases() -> void:
 				"unset_flag": _process_unset_flag_statement(args)
 
 
-
+## Processes a default_phrase statement like 'default_phrase dog_breed Labrador Retriever'
+##
+## Assigns the specified value to the specified chat history phrase if the phrase is currently unset.
+##
+## Parameters:
+## 	'args': The statement's arguments, such as ['dog_breed', 'Labrador' 'Retriever']
 func _process_default_phrase_statement(args: Array) -> void:
 	if args.size() < 2:
 		push_warning("Invalid argument count for default_phrase call. Expected at least 2 but was %s"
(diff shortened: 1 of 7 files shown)


project/src/demo/world/CutsceneDemo.tscn
```diff
@@ -165,16 +165,16 @@ dialog_autowrap = true
...
-[connection signal="about_to_show" from="Dialogs/OpenFile" to="Dialogs/Backdrop" method="_on_Dialog_about_to_show"]
-[connection signal="file_selected" from="Dialogs/OpenFile" to="." method="_on_OpenFileDialog_file_selected"]
-[connection signal="hide" from="Dialogs/OpenFile" to="." method="_on_OpenFile_hide"]
-[connection signal="popup_hide" from="Dialogs/OpenFile" to="Dialogs/Backdrop" method="_on_Dialog_popup_hide"]
+[connection signal="toggled" from="Input/LongNames" to="Input/LongNames" method="_on_toggled"]
 [connection signal="pressed" from="Input/Fatness/CheckBox" to="Input/Fatness" method="_on_CheckBox_pressed"]
 [connection signal="value_changed" from="Input/Fatness/HSlider" to="Input/Fatness" method="_on_HSlider_value_changed"]
-[connection signal="toggled" from="Input/LongNames" to="Input/LongNames" method="_on_toggled"]
 [connection signal="valid_changed" from="Input/Open" to="." method="_on_Open_valid_changed"]
 [connection signal="pressed" from="Input/Open/Button" to="." method="_on_OpenButton_pressed"]
```
Poobslag, line 172: rv

Poobslag: Reverted.

PR #1800, opened 2022-11-25
Removed yield statements from PuzzleState.

Poobslag:
Instead of yield statements. PuzzleState now uses timers and listeners to handle brief pauses. This prevents bugs caused by yield statements terminating at weird times, such as restarting a puzzle while a countdown is happening, or restarting a puzzle before the victory screen appears.

Changed puzzle-state.gd from a global script into a global node, to allow for a child 'Timers' node.

Moved tutorial Timers node into PuzzleState.


project/project.godot
```diff
@@ -1601,6 +1601,12 @@ move_piece_right={
 , Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":0,"alt":false,"shift":false,"control":false,"meta":false,"command":false,"pressed":false,"scancode":16777356,"physical_scancode":0,"unicode":0,"echo":false,"script":null)
  ]
 }
+retry={
+"deadzone": 0.5,
+"events": [ Object(InputEventKey,"resource_local_to_scene":false,"resource_name":"","device":0,"alt":false,"shift":false,"control":false,"meta":false,"command":false,"pressed":false,"scancode":82,"physical_scancode":0,"unicode":0,"echo":false,"script":null)
+, Object(InputEventJoypadButton,"resource_local_to_scene":false,"resource_name":"","device":0,"button_index":10,"pressure":0.0,"pressed":false,"script":null)
+ ]
+}
```
Poobslag, line 1609: rv

project/src/main/puzzle/puzzle-state.gd
```diff
@@ -112,20 +117,56 @@ func _physics_process(_delta: float) -> void:
...
+func add_timer(wait_time: float) -> Timer:
+	var timer := Timer.new()
+	timer.one_shot = true
+	timer.wait_time = wait_time
+	timer.connect("timeout", self, "_on_Timer_timeout_queue_free", [timer])
+	_timers.add_child(timer)
+	return timer
+
+
+## Cleans up any temporary timers and listeners.
```
Poobslag, line 153: Rename this method, it only cleans up timers.

Poobslag: I prefer the current name; all timer have associated listeners, and it's also symmetrical with the tutorial method of the same name.

project/project.godot
Poobslag: Removed.

PR #1815, opened 2022-11-30
PieceManager refactoring: initialize_piece, prespawn top out

Poobslag:
Extracted PieceManager._initialize_piece() utility method. This avoids bugs from duplicated code; some of these pieces were incorrectly initialized with the piece's tilemap which could cause complications down the road.

Prespawn state no longer explicitly changes to top out state. This simplifies the contract of PieceManager.spawn_piece(). We transition to the top out state anyways, so this logic was redundant.

### DIFF
--- a/project/src/main/puzzle/piece/piece-manager.gd
+++ b/project/src/main/puzzle/piece/piece-manager.gd
@@ -81,8 +81,8 @@ func _ready() -> void:
 	PuzzleState.connect("game_ended", self, "_on_PuzzleState_game_ended")
 	Pauser.connect("paused_changed", self, "_on_Pauser_paused_changed")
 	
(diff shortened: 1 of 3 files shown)


project/src/main/puzzle/piece/piece-manager.gd
```diff
@@ -139,15 +139,22 @@ func exit_top_out_state() -> void:
 
 ## Spawns a new piece at the top of the playfield.
 ##
-## Returns 'true' if the piece was spawned successfully, or 'false' if the player topped out.
-func spawn_piece() -> bool:
+## This is usually from the piece queue, but can also be the player's hold piece if they have the appropriate cheat
+## enabled and are holding the swap key.
```
Poobslag, line 143: Remove mention of hold piece.

Poobslag: Removed.

PR #1819, opened 2022-11-30
Added hold piece cheat. Changed limited piece level behavior.

Poobslag:
The player can enable a hold piece cheat via the Settings -> Gameplay menu.

If the player is using Guideline/WASD controls we define a key for them. If the player is using custom controls, a 'Swap Hold Piece' menu option is shown and hidden dynamically. We don't want to confuse players by letting them set a 'Swap Hold Piece' key which doesn't work because they didn't enable the cheat, which is why we hide it.

Limited piece levels now count the pieces spawned, not the pieces placed. Adding a 'hold piece' allows the player to skip pieces and not place them, resulting in unexpected behavior when the player only places 79/80 pieces but has no more pieces to place. So a level that requires 80 pieces can now be cleared by placing 79 pieces and swapping.


project/src/main/puzzle/Playfield.tscn
```diff
@@ -118,6 +118,9 @@ viewport_path = NodePath("TileMapClip/TileMap/GhostPieceViewport")
 margin_right = 324.0
 margin_bottom = 544.0
 script = ExtResource( 1 )
+__meta__ = {
+"_edit_use_anchors_": false
+}
```
Poobslag, line 123: rv

project/src/main/ui/settings/SettingsMenu.tscn
```diff
@@ -264,7 +265,6 @@ size_flags_vertical = 4
 size_flags_stretch_ratio = 1.1
 
 [node name="ScrollContainer" type="ScrollContainer" parent="Window/UiArea/TabContainer"]
-visible = false
```
Poobslag, line 267: why did this visibility toggle?

project/src/main/puzzle/piece/piece-queue.gd
```diff
@@ -72,6 +76,12 @@ func get_next_piece(index: int) -> NextPiece:
 	return pieces[index]
 
 
+func swap_hold_piece(piece_type: PieceType) -> NextPiece:
+	var old_hold_piece = hold_piece
```
Poobslag, line 80: Add type hint

project/src/main/puzzle/Playfield.tscn
Poobslag: Reverted

project/src/main/ui/settings/SettingsMenu.tscn
Poobslag: Fixed

project/src/main/puzzle/piece/piece-queue.gd
Poobslag: Changed to `:=`

PR #1826, opened 2022-12-07
Workaround for Stack Overflow in _on_resized()

### LINKED ISSUE #1810: Stack Overflow in combo-score-value-label.gd _on_resized()
image

Setting the `margin_left` can apparently trigger a resize event in some cases. This happened when playing the tutorial and trying to give up.

Poobslag:
combo-score-value-label.gd and score-value-label.gd could occasionally cause a Stack Overflow error, where assigning their margins would repeatedly change their rect size over and over, to a larger and larger negative value. This usually stabilized, but sometimes it would just go further and further negative forever. I noticed this mostly in Tutorials when I was using the debugger.

These labels now disconnect their listener so they cannot react to their own signals.

Closes #1810.

### DIFF
--- a/project/src/main/puzzle/score-value-label.gd
+++ b/project/src/main/puzzle/score-value-label.gd
@@ -17,7 +17,11 @@ func _on_resized() -> void:
 	if not _top_out_particles:
 		return
 	
+	# Avoid a Stack Overflow where changing our margins triggers another _on_resized() event, see #1810
+	disconnect("resized", self, "_on_resized")
 	margin_left = -rect_size.x
+	connect("resized", self, "_on_resized")
+	
 	_top_out_particles.rect_position = rect_size * 0.5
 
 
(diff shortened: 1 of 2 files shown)


project/src/main/puzzle/score-value-label.gd
```diff
@@ -17,8 +17,11 @@ func _on_resized() -> void:
 	if not _top_out_particles:
 		return
 	
+	# Avoid a Stack Overflow where changing our margins triggers another _on_resized() event, see #1810
+	disconnect("resized", self, "_on_resized")
 	margin_left = -rect_size.x
 	_top_out_particles.rect_position = rect_size * 0.5
+	connect("resized", self, "_on_resized")
```
Poobslag, line 24: Maybe move this line up for symmetry with the other similar code. There is no risk of this rect_position call triggering an on_resized event.

Poobslag: Moved.

PR #1864, opened 2022-12-13
Low priority resource paths: CareerMap, CareerWin...

Poobslag:
Added more low priority resource paths. These paths were taking >100 ms to load on my PC, but loading them later makes the loading screen run more smoothly.

### DIFF
--- a/project/src/main/utils/ResourceCache.tscn
+++ b/project/src/main/utils/ResourceCache.tscn
@@ -4,6 +4,6 @@
 
 [node name="ResourceCache" type="Node"]
 script = ExtResource( 1 )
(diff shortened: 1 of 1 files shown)


Poobslag:
The goal of this merge was to make the loading screen less jittery.

I experimented with [splitting up Puzzle.tscn](https://github.com/Poobslag/turbofat/files/10219927/out.txt) which reduced the jitter of Puzzle.tscn from 170 ms to 120 ms, but it makes the design much uglier to me. Maybe there's a more elegant way to extract this which we can consider some day -- something along the lines of a raw bare-bones puzzle scene with 'decorators' around it, or something like that. Each decorator could be its own scene, and they could be modular.

PR #1869, opened 2022-12-14
Careers. Resolved CareerData/CareerRegion cyclic reference.

Poobslag:
Extracted new 'Careers' constants class.

Fixed cyclic reference between CareerData and CareerRegion which was introduced in 8c0a0450

### DIFF
--- a/project/project.godot
+++ b/project/project.godot
@@ -84,6 +84,11 @@ _global_script_classes=[ {
 "language": "GDScript",
 "path": "res://src/main/career/career-rewinder.gd"
 }, {
+"base": "Reference",
+"class": "Careers",
+"language": "GDScript",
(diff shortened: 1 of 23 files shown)


project/src/main/puzzle/FoodItem.tscn
```diff
@@ -65,6 +65,8 @@ texture = ExtResource( 2 )
 frame_data = "res://assets/main/puzzle/food/food-packed.json"
 rect_size = Vector2( 128, 128 )
 frame = 1
+scale_modifier = Vector2( 1, 1 )
+rotation_modifier = 0.0
```
Poobslag, line 69: rv

Poobslag: Reverted.

PR #1870, opened 2022-12-14
Extracted 'is_tutorial' utility method

Poobslag:


project/src/main/puzzle/hold-piece-displays.gd
```diff
@@ -28,7 +28,7 @@ func _ready() -> void:
 
 
 func _refresh() -> void:
-	visible = CurrentLevel.hold_piece_enabled()
+	visible = CurrentLevel.is_hold_piece_cheat_enabled()
```
Poobslag, line 31: What is this behavior like after a tutorial?

Poobslag: Not good. I've fixed it with 6f514a71d.

PR #1875, opened 2022-12-16
ConfigStringUtils.config_string_from_ints() now generates ellipses

Poobslag:
config_string_from_ints used to not understand patterns like '11,12,13,14,15,16...' and would generate a string with 20,000 characters. It now generates an ellipsis string in the same format that we use for input.


project/src/main/puzzle/level/config-string-utils.gd
```diff
@@ -3,42 +3,78 @@ class_name ConfigStringUtils
...
+				
+				if i < ints.size() - 1:
+					range_start = ints[i + 1]
+	
+	# output an ellipse expression if necessary
+	if increment == 0:
+		# no sequence detected; do not output an ellipsis expression
+		pass
+	else:
+		# produce output like '6,8...'
```
Poobslag, line 69: Change this comment, it's not quite accurate

Poobslag: Removed; comment was redundant an inaccurate

PR #1880, opened 2022-12-19
LineFiller can now leave empty lines.

Poobslag:
Instead of filling ALL empty lines, LineFiller now only fills the top few.

This opens the possibility for levels which fill in pickups every 3 rows, but leave the other 2 rows empty.

This causes a potential negative interaction with levels like 'above and below' which insert lines in the middle of the playfield as a part of a line clear. Our new LineFiller algorithm will only look at the top 3 rows, so it will ignore empty lines in the middle of the playfield. This seems OK -- if a level like that wanted to have 'filled lines' in the middle of the playfield it could insert them, instead of inserting empty lines.

### DIFF
--- a/project/src/main/puzzle/line-filler.gd
+++ b/project/src/main/puzzle/line-filler.gd
@@ -85,14 +85,21 @@ func _schedule_playfield_refill() -> void:
 		_lines_being_filled.append(max(1, (visible_row_count - i) * per_line_fill_delay))
 
 
-## Immediately fills all empty lines on the playfield.
-func _fill_empty_lines() -> void:
+## Immediately fills empty lines on the playfield.
+##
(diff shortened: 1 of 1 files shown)


project/src/main/puzzle/line-filler.gd
```diff
@@ -85,14 +85,14 @@ func _schedule_playfield_refill() -> void:
 		_lines_being_filled.append(max(1, (visible_row_count - i) * per_line_fill_delay))
 
 
-## Immediately fills all empty lines on the playfield.
-func _fill_empty_lines() -> void:
+## Immediately fills empty lines on the playfield.
+func _fill_empty_lines(filled_line_count: int) -> void:
```
Poobslag, line 89: Document this new parameter.

PR #1908, opened 2023-01-23
'Welcome To Turbofat' boss level advances player to the next region.

Poobslag:
This is redundant, but consistent with the 'Lemony Thickets' boss level.


Poobslag:
It makes sense that all the cutscenes set in the "walking" environment should advance the cutscene, it's intuitive and prevents dumb oversights where we have them walking somewhere and then they're in the same region afterwards.

PR #1923, opened 2023-02-01
Fixed bad spawn points in undecorated environment

Poobslag:
This made characters show up in the wrong position for some cutscenes.


Poobslag:
image

Here is an example where the player spawned in the upper right corner, so Tunathy is turned away from them instead of talking toward them.

PR #1969, opened 2023-03-08
push_back() -> append()

Poobslag:
Differen places were calling queue.push_back() or queue.append(), including PieceQueue which was alternating between both. I don't really care which we use, but the inconsistency was bothering me


Poobslag:
Intuitively I think 'append' makes more sense, since we should only be appending to the end of a queue. Pushing makes more sense for a stack. It avoids accidental calls to 'queue.push_front'.

PR #1970, opened 2023-03-08
Renamed restaurant creature_index -> customer_index

Poobslag:
'customer_index' is more specific. There are creatures in the scene that are not customers (the chef).

### DIFF
--- a/project/src/main/puzzle/restaurant-view.gd
+++ b/project/src/main/puzzle/restaurant-view.gd
@@ -17,7 +17,7 @@ const MOUTH_POSITIONS_BY_ORIENTATION := {
 const SWOOP_DURATION := 0.8
 
 ## virtual property; value is only exposed through getters/setters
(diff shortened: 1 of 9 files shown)


project/src/main/puzzle/restaurant-view.gd
```diff
@@ -28,10 +28,10 @@ var _customer_offscreen_rect_position: Vector2
 var _chef_onscreen_rect_position: Vector2
 var _chef_offscreen_rect_position: Vector2
 
-## Ordered list of int indexes which will be cycled to with the next_creature_index or scroll_to_new_creature methods.
+## Ordered list of int indexes which will be cycled to with the next_customer_index or scroll_to_new_creature methods.
```
Poobslag, line 31: Should this be 'scroll_to_new_customer'?

Poobslag: Changed to `scroll_to_new_customer`. Also changed many other similar methods from `creature` to `customer`.

PR #1973, opened 2023-03-08
Removed goop_texture_offset shader parameter.

Poobslag:
This shader parameter dates back to when the shader was authored in June 2020, but the parameter has never done anything.


Poobslag:
My best guess is that the four walls were rendered individually during development, and this `goop_texture_offset` was assigned to each one. Before the code was merged, I made this one big wall to make things easier, and this texture offset was never implemented.

PR #1983, opened 2023-03-13
Moved exits to ground layer, instead of z_index=1

Poobslag:
OverworldExit used to have a z_index of -1, causing them to be rendered behind other nodes. This is fine but it contradicts the approach we have for other floor nodes, which are all placed in their own layer.

I've removed OverworldExit's z_index and moved them into the 'ground' node.


project/src/demo/world/environment/lemon/LemonFreeRoamEnvironment.tscn
```diff
@@ -71,6 +72,29 @@ pebble_density = 0.015
 ground_no_goop_tile_index = 1
 pebble_tile_index = 1
 
+[node name="Exits" type="Node2D" parent="Ground"]
```
Poobslag, line 75: Document the inclusion of these new exits.

Poobslag: Expanded commit comment, `Added missing LemonFreeRoamEnvironment exits.`

PR #2000, opened 2023-04-05
Squish drops now emit a hard drop signal.

Poobslag:


Poobslag:
This behavior is important for the upcoming sharks who can be squished by hard drops. They need to be squished by squish drops as well, but currently there is no way to detect squish drops.

PR #2022, opened 2023-04-17
Fixed bug with sharks deleting during multiline inserts

Poobslag:
'sharks for playfield' are now refreshed once after all lines are deleted/inserted, rather than refreshing them for every inserted lines.

Playfield.is_clearing_lines() now returns true throughout LineClearer._delete_lines() method. Before, is_clearing_lines() returned false while triggers were firing, meaning bugs related to inserted/deleted lines were really difficult to work around.

### DIFF
--- a/project/src/main/puzzle/line-clearer.gd
+++ b/project/src/main/puzzle/line-clearer.gd
@@ -108,13 +108,15 @@ func _physics_process(_delta: float) -> void:
 		var old_lines_being_erased := lines_being_erased
 		var old_lines_being_deleted := lines_being_deleted
 		
-		set_physics_process(false)
-		
 		lines_being_cleared = []
 		lines_being_erased = []
 		lines_being_deleted = []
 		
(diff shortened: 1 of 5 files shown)


project/src/main/puzzle/line-clearer.gd
```diff
@@ -108,13 +108,13 @@ func _physics_process(_delta: float) -> void:
...
 		
-		set_physics_process(false)
-		
 		lines_being_cleared = []
 		lines_being_erased = []
 		lines_being_deleted = []
 		
 		_delete_lines(old_lines_being_cleared, old_lines_being_erased, old_lines_being_deleted)
+		
+		set_physics_process(false)
```
Poobslag, line 117: 'is_clearing_lines' will return true during this entire process, but 'is_box_building' will return false during the final signal. For consistency, we should think more carefully about when we call 'set_physics_process(false)' and add a comment.

Poobslag: Added comment 'causes 'Playfield.is_building_boxes()' to return false...' and modified BoxBuilder for consistency.

PR #2082, opened 2023-07-12
Refactor: queue_free when calling remove_child

Poobslag:


Poobslag:
I've also reordered a few existing `queue_free()` calls for consistency. It seems logical to remove() and then queue_free() after. Sometimes we want to remove a node from its parent without knowing who its parent is. If we call queue_free() before remove() then we're dependent on the node temporarily preserving its state for the get_parent() call, and even if that works I wouldn't want to depend on any of the node's properties after a queue_free call.

PR #2153, opened 2023-09-01
Newly spawned pieces can slide twice during 20G.

### LINKED ISSUE #2136: Newly spawned pieces don't always slide to fill certain holes
https://github.com/Poobslag/turbofat/assets/5902413/360b6621-2edb-4079-b36f-963832259efc

Per Hish from the Turbo Fat discord:

> At high speed, if I hold a correct direction and rotation, some parts don't fit in some holes, like in the first example (Sorry if it's too fast), the piece can't "jump" one piece up. It can be corrected by rotating the piece so it can make a box, as in the second example. But this doesn't happen when you squish a piece from the top of that hole, this piece can "jump", as in the last example, but a new spawned piece can't. It's kind of inconsistent and hard to react at high speed, or maybe it's just me

Initial DAS should be able to slide pieces into gaps like this.

Poobslag:
Instead of mid-drop movement sliding the piece only once, it will slide the piece up to 8 times if DAS is active.

Closes #2136.

### DIFF
--- a/project/src/main/puzzle/piece/piece-mover.gd
+++ b/project/src/main/puzzle/piece/piece-mover.gd
@@ -37,14 +37,26 @@ const SPAWN_RIGHT := [
 
 export (NodePath) var input_path: NodePath
 
-## how many times the piece has moved horizontally this frame
-var _horizontal_movement_count := 0
+## how many times the piece has moved horizontally this frame, as a result of normal movement (not mid-drop movement)
+var _normal_horizontal_movement_count := 0
 
 onready var input: PieceInput = get_node(input_path)
 
 func _physics_process(_delta: float) -> void:
-	_horizontal_movement_count = 0
+	_normal_horizontal_movement_count = 0
 
+
+## Applies initial move input to the specified piece.
+##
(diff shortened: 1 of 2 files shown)


project/src/main/puzzle/piece/piece-mover.gd
```diff
@@ -120,9 +151,23 @@ func apply_move_input(piece: ActivePiece) -> void:
...
-	if _horizontal_movement_count == 0:
-		apply_move_input(piece)
+	if _normal_horizontal_movement_count > 0:
+		# piece moved normally this frame, do not apply mid-drop movement
+		return
+	
+	# attempt a single mid-drop movement
+	var movement_signal := apply_move_input(piece, true)
+	
+	if movement_signal != "" and (input.is_left_das_active() or input.is_right_das_active()):
```
Poobslag, line 167: Couldn't this check for the movement signal's value of `das_moved_left` to be a little more concise?

Poobslag: Updated code to check the movement signal instead of checking `input` directly

PR #2154, opened 2023-09-01
Moving no longer interrupts mood. Disabled north-facing emotes.

### LINKED ISSUE #1905: Heart eyes appear on the back of their head when facing backwards
AYO_FLUTTER_WT

If a character does the 'heart eye' emote when facing backwards, the heart eyes appear on the back of their head.

Poobslag:
Instead of interrupting the movement animations, moving just resets the creature's "emote arm frame". This is the same mechanism that happens when you launch an emote while the creature is running.

Creatures will no longer emote when facing north. This was resulting in some buggy behavior like eyes on the back of their head.

Closes #1905.

Added movement options to MoodDemo.

### DIFF
--- a/project/src/demo/chat/ui/mood-demo.gd
+++ b/project/src/demo/chat/ui/mood-demo.gd
@@ -19,6 +19,7 @@ extends Node
 ## 	[space bar]: Feed
 ## 	[brace keys]: Change the creature's appearance
 ## 	[Shift + /]: Print DNA
+## 	[right]: Toggle creature movement
 ##
 ## 	[Shift + Q, W]: Look over shoulder
 ## 	[Shift + E, R]: Yawn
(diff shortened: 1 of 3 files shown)


project/src/demo/chat/ui/mood-demo.gd
```diff
@@ -41,6 +42,10 @@ func _ready() -> void:
 		_creature.creature_def = CreatureDef.new().from_json_path(creature_path)
 
 
+func _physics_process(delta: float) -> void:
```
Poobslag, line 45: This is a little strange; add a comment

Poobslag: Additionally, `delta` is not used in this function and should be prefixed with an underscore.

Poobslag: Added comment, `# The creature movement toggle actually moves the creature....`

PR #2174, opened 2023-09-06
Combined 'flawed tilers' like CarpetTiler into one script

Poobslag:
CarpetTiler, FlawedTiler and KitchenFloorTiler all had essentially the same script.


project/src/main/world/environment/restaurant/flawed-tiler.gd
```diff
@@ -0,0 +1,105 @@
+tool
```
Poobslag, line 1: This script is in the wrong place, as it is not restaurant-specific

PR #2193, opened 2023-09-29
Added speedups to a few levels.

Poobslag:
It's fun when levels get harder over time, but not many levels do it. I even made the first level where each customer makes the level get a little faster, 'midnight_snack'.

### DIFF
--- a/project/assets/main/puzzle/levels/career/a-little-garbage.json
+++ b/project/assets/main/puzzle/levels/career/a-little-garbage.json
@@ -1,12 +1,19 @@
 {
   "version": "49db",
-  "start_speed": "1",
+  "start_speed": "4",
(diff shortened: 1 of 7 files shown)


project/assets/main/puzzle/levels/career/a-little-garbage.json
```diff
@@ -1,12 +1,19 @@
 {
   "version": "49db",
-  "start_speed": "1",
+  "start_speed": "4",
```
Poobslag, line 3: Document these speed adjustments.

Poobslag: Added commit message, `Adjusted a few level speeds to fit their corresponding career region.`

PR #2194, opened 2023-09-29
Retired old levels from Cannoli Sandbar, later

Poobslag:
The retired levels were either not very fun to me, or included gimmicks which other levels did better (like Creeping Crevice versus Creeping Crevice 2.) Veggie Haters and Wedding Cake For One had an additional problem where a typical player might score $50-$100, but a skilled player could score $3,000 or more if they were extremely dedicated.

Overall these levels were retired because I'd prefer to have 100 great levels rather than 200 mediocre ones.


project/src/main/data/player-save-upgrader.gd
```diff
@@ -290,7 +321,7 @@ func _upgrade_38c3(_old_save_items: Array, save_item: SaveItem) -> SaveItem:
 	return save_item
 
 
-func _upgrade_37b3(_old_save_items: Array, save_item: SaveItem) -> SaveItem:
+func _upgrade_3776(_old_save_items: Array, save_item: SaveItem) -> SaveItem:
```
Poobslag, line 324: Document this in the commit message.

Poobslag: Added commit message, `Fixed inconsistent naming of '_upgrade_37b3()' methods ...`

PR #2227, opened 2023-10-18
Chocolava: Paths, puddles, ripple sprites

Poobslag:
Added paths, puddles, pebbles and powdered sugar. The puddles are adapted from Poki Desert.


todo.txt
```diff
@@ -0,0 +1,3 @@
+-comments
```

PR #2288, opened 2023-12-27
Hardcore levels.

Poobslag:
The player is offered a hardcore level at least twice each career session. Hardcore levels are regular levels, but the player only gets one life. If they fail, they have to start the career session over. If they succeed, they get an extra life.

The hardcore level select buttons have a very unique appearance. Unfortunately Godot does not have an easy way to layer Controls with sprites, because Controls lack a Z-index property. To work around this limitation I made a lot of really ugly changes, including changing LevelSelectButton from a Button to a Control, and adding a few sprites to the scene tree for the sole purpose of giving Controls a necessary z coordinate. I also changed the HookableLevelGradeLabel scene's z coordinate so it appears over top of the skull effect.


project/src/main/career/career-data.gd
```diff
@@ -15,6 +15,13 @@ signal distance_travelled_changed
 ## mid-scene, and really only happens when using cheat codes.
 signal hours_passed_changed
 
+const HARDCORE_LEVEL_HOURS_OPTIONS := [
```
Poobslag, line 18: Add a comment

Poobslag: Added comment, `...the hours_passed values where ...`

PR #2302, opened 2024-01-03
region_cleared -> region_finished

Poobslag:
These are synonyms, but this new term is consistent with the 'level_finished' flag.


Poobslag:
After reviewing this pull request, I've opened #2303 because I forgot to add these `customer_if` flags to customers in the newer areas.

PR #2306, opened 2024-01-03
Smaller shadow tiles.

Poobslag:
Shadow tiles used to be 144x128 which was almost twice as large as they needed to be. This made some of the sprite sheets over 2000x2000 pixels in size, which impacts performance and load times.


Poobslag:
I initially wanted to avoid the repetitive shadow sprite sheets and have them just be one or two rows. But I couldn't think of how to do this without introducing some mapping logic. I think the current approach is pretty braindead and simple which is sometimes a good thing.

PR #2310, opened 2024-01-05
Remove 'daily' from career fields.

Poobslag:
This was haphazardly applied. Some fields like 'distance_travelled' were not called 'daily_distance_travelled' even though they were properties of the current career day. Other fields like 'daily_steps' had the word 'daily'.

I've removed the word 'daily' from all career fields, and updated our backwards compatibility logic to remove it from the player's saves as well.

### DIFF
--- a/project/src/main/career/career-calendar.gd
+++ b/project/src/main/career/career-calendar.gd
@@ -65,9 +65,9 @@ func advance_clock(new_distance_earned: int, success: bool, lost: bool) -> void:
 
 ## Advances the calendar day and resets all daily variables.
 func advance_calendar() -> void:
-	career_data.prev_daily_earnings.push_front(career_data.daily_earnings)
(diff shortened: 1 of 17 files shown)


project/src/main/career/career-calendar.gd
```diff
@@ -77,11 +77,11 @@ func advance_calendar() -> void:
...
-	career_data.daily_customers = 0
-	career_data.daily_earnings = 0
-	career_data.daily_level_ids.clear()
-	career_data.daily_seconds_played = 0.0
-	career_data.daily_steps = 0
+	career_data.customers = 0
+	career_data.money = 0
+	career_data.level_ids.clear()
+	career_data.seconds_played = 0.0
+	career_data.steps = 0
```
Poobslag, line 84: Let's alphabetize this, similar code is alphabetized elsewhere.

Poobslag: Alphabetized.

PR #2311, opened 2024-01-05
*_environment_name -> *_environment_id

Poobslag:
Fields like 'overworld_environment_name' are now called 'overworld_environment_id' instead. I think this makes more sense because these are unique identifiers like 'lava/zagma' and not names like 'Chocolava Canyon'.


project/src/main/world/career-world.gd
```diff
@@ -159,8 +159,8 @@ func _remove_mile_markers() -> void:
 
 
 func _career_environment_path() -> String:
-	var environment_name := PlayerData.career.current_region().overworld_environment_name
-	return ENVIRONMENT_PATH_BY_NAME.get(environment_name, DEFAULT_ENVIRONMENT_PATH)
+	var environment_name := PlayerData.career.current_region().overworld_environment_id
```
Poobslag, line 162: Change this variable name to `environment_id`

Poobslag: Changed.

PR #2320, opened 2024-01-26
Added credits scrolling logic.

Poobslag:
The credits are scriptable, which is handled through SceneTreeTweens. For now, these events just involve basic 'move left' 'hide the title' type layout changes since those would massively affect how the scene was organized.


project/src/main/credits/credits.gd
```diff
@@ -0,0 +1,48 @@
...
+enum CreditsPosition {
+	TOP_LEFT,
+	TOP_RIGHT,
+	BOTTOM_LEFT,
+	BOTTOM_RIGHT,
+	CENTER_TOP,
+	CENTER_BOTTOM,
+}
+
+const FADE_RANGE := 10
```
Poobslag, line 17: Document this

Poobslag: Added comment, ` ## The distance a credits line will travel...`

PR #2331, opened 2024-02-21
Viewport: size_override_stretch=true, hdr=false

Poobslag:
Some of our viewports use subtly different settings, so I'm making them more consistent. I also notice a few use different sound settings, but I'm leaving this alone for now. I think they do it to avoid having sound effects play in the left or right channel in weird ways.

### DIFF
--- a/project/src/main/credits/CreditsScroll.tscn
+++ b/project/src/main/credits/CreditsScroll.tscn
@@ -1,4 +1,4 @@
-[gd_scene load_steps=19 format=2]
+[gd_scene load_steps=20 format=2]
 
 [ext_resource path="res://src/main/credits/CreditsHeader.tscn" type="PackedScene" id=1]
 [ext_resource path="res://assets/main/ui/menu/loading-orb-sheet.png" type="Texture" id=2]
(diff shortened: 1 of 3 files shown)


project/src/main/world/environment/restaurant/RestaurantPuzzleScene.tscn
```diff
@@ -16,11 +14,7 @@ script = ExtResource( 3 )
 [node name="World" type="Node" parent="."]
 script = ExtResource( 2 )
 
-[node name="Environment" type="Node2D" parent="World" groups=["overworld_environments"] instance=ExtResource( 4 )]
-script = ExtResource( 6 )
-environment_shadows_path = NodePath("Ground/Shadows")
-obstacles_path = NodePath("Obstacles")
-CreatureScene = ExtResource( 7 )
+[node name="Environment" parent="World" instance=ExtResource( 4 )]
```
Poobslag, line 17: rv

Poobslag: Reverted

PR #2338, opened 2024-02-21
credits-scroll.gd warning. CutsceneCamera refactoring.

Poobslag:
Fixed 'unused parameter' warning in credits-scroll.gd.

CutsceneCamera now iterates over _overworld_ui.chatters instead of all creatures. This makes it consistent with similar methods in OverworldUi.


project/src/main/world/cutscene-camera.gd
```diff
@@ -37,14 +37,14 @@ func snap_into_position() -> void:
 func _calculate_camera_bounding_box() -> Rect2:
 	return _overworld_ui.get_chatter_bounding_box().grow(CAMERA_BOUNDARY)
 
-
```
Poobslag, line 40: rv

Poobslag: Reverted

PR #2360, opened 2024-03-05
Crowd -> Crowdie

Poobslag:
There's no english word for 'person within a crowd', but calling each crowd member as a 'crowd' results in confusing and ambiguous nomenclature. I've settled on the word 'crowdie'.


project/src/main/world/environment/lava/crowdie-recycler.gd
Poobslag, line 1: This should have some renames and stuff.

Poobslag: Updated comments throughout project, `This lets us reuse the same ~150 crowdies instead of creating ~1,500 crowdies...`

PR #2368, opened 2024-03-08
Removed redundant meta flags. Added delint check.

Poobslag:
Removed redundant meta flags: _edit_use_anchors=false, _editor_description="", _edit_vertical_guides=[]. A bug in the Godot editor cause meta flags such as '_edit_use_anchors=false' to randomly manifest, even when the properties are set to their default values. I've added a delint check to ensure these values don't pollute our scene files.

Workaround for Godot #57862 (https://github.com/godotengine/godot/issues/57862)


project/src/main/editor/creature/CreatureEditor.tscn
```diff
@@ -1299,10 +1295,7 @@ margin_bottom = 595.0
 theme = ExtResource( 2 )
 shortcut_in_tooltip = false
 text = "Quit"
-__meta__ = {
-"_edit_use_anchors_": false,
-"_editor_description_": ""
-}
+
```
Poobslag, line 1298: rv

project/src/main/ui/FontFitButton.tscn
```diff
@@ -7,6 +7,4 @@ margin_right = 12.0
 margin_bottom = 20.0
 clip_text = true
 script = ExtResource( 1 )
-__meta__ = {
-"_editor_description_": ""
-}
+
```
Poobslag, line 10: rv

project/src/main/editor/creature/CreatureEditor.tscn
Poobslag: Reverted

PR #2382, opened 2024-03-26
Added credits pinup transformations.

Poobslag:
Orb now fires some of its pieces towards pinups, which are then transformed into silly anime caricatures of the characters.

Tweaked AnimationPlayer sync; Syncing the AnimationPlayer to 4.100 FPS works better with the music, which is 123 beats per minute. It's still not perfectly in sync with the music but it's as close as I think I can get.


project/src/main/ui/menu/LoadingPiece.tscn
```diff
@@ -14,13 +14,13 @@ particles_anim_loop = false
 offsets = PoolRealArray( 0, 0.665217, 1 )
 colors = PoolColorArray( 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 0 )
 
-[sub_resource type="GradientTexture" id=3]
```
Poobslag, line 17: rv

Poobslag: reverted

PR #2387, opened 2024-04-03
Crowd cutscene collision bugs. Added 'stop'.

Poobslag:
Fixed crowd cutscene collision bugs. CrowdSurfEnvironment and CrowdWalkEnvironment both use a CrowdRecycler to shuffle creatures around when they collide with an Area2D. However in fabd4cc8 we started loading both these environments simultaneously to reduce lag. This had the unintended side effect of these two Area2Ds interacting with creatures in the other environment.

I've fixed the bug by disabling these Area2D collision areas when they are not in use. Cutscenes are now considered active when 'play' is called, and they are considered inactive when a new 'stop' method is called. Inactive cutscenes have all their animations stopped and their CrowdRecyclers disabled.

This new 'stop' method is also available in CrowdWalkCutsceneDemo and CrowdSurfCutsceneDemo.

### DIFF
--- a/project/src/main/world/environment/lava/crowd-walk-director.gd
+++ b/project/src/main/world/environment/lava/crowd-walk-director.gd
@@ -7,6 +7,9 @@ extends Node
 ## Emitted when creatures are moved to their starting positions for the cutscene.
 signal played
 
+## Emitted when the cutscenes animations are stopped.
+signal stopped
+
 ## Number in the range [0.0, 1.0] for how many creatures should jump up and down with their arms raised.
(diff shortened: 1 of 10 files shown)


project/src/main/world/environment/lava/crowd-walk-director.gd
```diff
@@ -6,6 +6,7 @@ extends Node
 
 ## Emitted when creatures are moved to their starting positions for the cutscene.
 signal played
+signal stopped
```
Poobslag, line 9: ## Emitted when the cutscenes animations are stopped.

Poobslag: Added missing comment.

PR #2404, opened 2024-04-17
Revert "Changed export presets to zips. Linux +x flag."

Poobslag:
Reverts Poobslag/turbofat#2399


Poobslag:
Godot 3.x does not support exporting .zip files, this feature is new to Godot 4.x.

PR #2407, opened 2024-04-17
Enable +x flag for linux executable, using 'zip_exec' utility

Poobslag:
It is difficult in Windows to create zip files with the executable flag set. Godot 4.x has support for exporting directly as zips which solves this problem. For Godot 3.x, we have to solve this problem ourselves.


bin/zip_exec/readme.txt
```diff
@@ -0,0 +1,23 @@
...
+zip_exec.exe "c:\path to zip\file.zip" "some path/some file in the zip"
+
+
+---
+Note also that to make it work on both mac and linux, unix attributes will be set on ALL files
+ and directories.
+
+Directories  drw-r--r--
+Executables  -rwxr-xr-x
+Normal files -rw-r--r--
```
Poobslag, line 23: Can you give this a pass and clean up the whitespace and formatting?

bin/zip_exec/readme.txt
```diff
@@ -0,0 +1,25 @@
...
+
+zip_exec.exe "C:\path_to_zip\file.zip" "some_path/some_file_in_the_zip.exe"
+
+## Note
+
+When using zip_exec on macOS or Linux, Unix attributes will be set on all files and directories within the zip archive:
+
+- Directories: drw-r--r--
+- Executables: -rwxr-xr-x
+- Normal files: -rw-r--r--
```
Poobslag, line 25: Align these flags so they're easier to compare

bin/zip_exec/readme.txt
Poobslag: I've cleaned up the whitespace and formatting based on some perceived unix standards and markdown.

bin/zip_exec/readme.txt
```diff
@@ -0,0 +1,23 @@
+
+A program to set the executable flag of a file within a zip archive.
+
+Useful in windows when you need to create zip files with valid
+ executables directly after unpacking on linux or mac.
+
+
+You may need to install vcredist_x86.exe to make it run.
+It can be downloaded from http://www.microsoft.com/download/en/details.aspx?displaylang=en&id=29
```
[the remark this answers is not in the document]
Poobslag: Updated to https://learn.microsoft.com/en-us/cpp/windows/latest-supported-vc-redist?view=msvc-170#visual-studio-2013-vc-120

PR #2463, opened 2024-05-06
Simplified 'Demo which shows off X' to 'Demonstrates X'

Poobslag:


project/src/demo/ui/menu/bad-save-data-control-demo.gd
```diff
@@ -1,5 +1,5 @@
 extends Node
-## Demo which shows off the popup window which is shown if there is a problem loading the save data
+## Demonstrates the popup window which is shown if there is a problem loading the save data
```
Poobslag, line 2: As long as we're fiddling with these comments, add the missing punctuation too

Poobslag: Changed to `...the save data.`, with an ending period

PR #2479, opened 2024-05-17
Added CandyButtonH4, a smaller CandyButton

Poobslag:


todo.txt
```diff
@@ -0,0 +1,21 @@
+*export one pattern
```

PR #2491, opened 2024-05-21
Removed Vega Churn Twelve. Moved levels to other regions.

Poobslag:


Poobslag:
Care was taken to move these levels to areas where the theme was appropriate. For example, "Three Course Meal" was moved to Poki Desert since it uses the night theme, and "Checkerboard Vault" was moved to Merrymellow Marsh because it is difficult.

The most notable exception is that Strawberry Shortcut has nothing to do with sharks, and as a result the concentration of shark-based levels in Poki Desert decreased. However, there is a pending update to address this.

PR #2517, opened 2024-05-31
SettingsMenu: Keybinds are now centered

Poobslag:
The keybinds used to be flush with the left side of the window, but they're now centered. This was a subtle issue as the window was almost the same size as the keybinds, but it will become more prominent when the window is enlarged.


project/src/main/ui/settings/SettingsMenu.tscn
```diff
@@ -1295,7 +1309,6 @@ dialog_text = "Change save slot?"
 dialog_autowrap = true
 
 [node name="DeleteConfirmation1" type="ConfirmationDialog" parent="Dialogs"]
-visible = true
```
Poobslag, line 1298: rv visibility of these dialogs

Poobslag: reverted

PR #2530, opened 2024-06-12
Extracted redundant gradient constants

Poobslag:
These gradients constants are currently used in two places, but the CreatureEditor will introduce several more. It's helpful to avoid the repetition. We have some proposed refactorings that will handle this in a cleaner way.

### DIFF
--- a/project/src/main/ui/candy-button/candy-button-h4.gd
+++ b/project/src/main/ui/candy-button/candy-button-h4.gd
@@ -16,50 +16,11 @@ export (Array, Font) var fonts := [
 	preload("res://src/main/ui/candy-button/candy-h6-font.tres"),
 ]
 
-## Gradient which colors the button bright cyan when the button is focused.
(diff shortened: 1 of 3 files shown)


project/src/main/ui/candy-button/candy-button-h4.gd
```diff
@@ -16,51 +16,12 @@ export (Array, Font) var fonts := [
 	preload("res://src/main/ui/candy-button/candy-h6-font.tres"),
 ]
 
-## Gradient which colors the button bright cyan when the button is focused.
-var _gradient_focused: Gradient = preload("res://src/main/ui/candy-button/gradient-focused.tres")
-
-## Gradient which colors the button bright cyan when the button is focused and hovered.
```
Poobslag, line 22: Preserve this comment.

Poobslag: Preserved comment;  `## Gradient which colors the button bright cyan when the button is focused and hovered.`

PR #2548, opened 2024-06-29
Redesigned creature editor for gamepads, non-default UI

Poobslag:
Redesigned the creature editor. The new design avoids use of the default grey Godot UI, and default widgets such as sliders and text boxes. It is also very gamepad friendly. It does not require scrolling, and categories can be navigated using the bumpers.

The old creature editor is still accessible via 'CreatureEditorOld' in the demo.

There is still some work remaining for the new creature editor:
 - name entry still requires keyboard
 - the color picker interface does not support gamepads
 - the old 'mutate' functionality was fun, and would be nice to keep around in some form
 - chat themes can still not be edited, and the constant randomization of them makes the 'body color picker' look a little strange as the nametag is constantly flickering


godot-l10n.pot
Poobslag, line 1: Remove.

project/src/demo/world/cutscene-demo.gd
```diff
@@ -76,7 +76,7 @@ func _on_OpenFileDialog_file_selected(path: String) -> void:
 	
 	if not _open_input.valid:
 		# show a dialog if the player selects an invalid cutscene
-		_error_dialog.dialog_text = "%s doesn't seem like the path to a cutscene file." % [path]
+		_error_dialog.dialog_text = tr("%s doesn't seem like the path to a cutscene file.") % [path]
```
Poobslag, line 79: rv

project/src/main/ui/focus-coordinator.gd
```diff
@@ -0,0 +1,127 @@
...
+## 		appropriate direction.
+##
+## 	3. Self. If no suitable focus neighbour is found, the control is assigned as its own focus neighbour. This prevents
+## 		Godot's default focus behavior from taking over.
+
+## Controls to manage focus neighbours for.
+var _controls: Array
+
+## Key: (Vector2) Unit vector for a cardinal direction.
+## Value: (String) Focus neighbour control property for the specified cardinal direction.
```
Poobslag, line 24: We don't usually capitalize these.

project/src/main/ui/tab-input-enabler.gd
```diff
@@ -6,7 +6,7 @@ extends Node
 
 ## Columns of focusable nodes below our TabContainer.
 ##
-## The first visible node of each of these arrays will have its its focus_neighbor_top assigned our TabContainer, so
+## The first visible node of each of these arrays will have its its focus_neighbour_top assigned our TabContainer, so
```
Poobslag, line 9: Move this into its own commit (sorry)

godot-l10n.pot
Poobslag: Removed.

project/src/demo/world/cutscene-demo.gd
Poobslag: Reverted.

project/src/main/ui/focus-coordinator.gd
Poobslag: Lower cased: `## key: (Vector2) Unit vector for a cardinal direction.`

project/src/main/ui/tab-input-enabler.gd
Poobslag: Pulled into #2549

PR #2560, opened 2024-07-16
Extracted candy button ClickSound, HoverSound scenes

Poobslag:
HoverSound was in 8 different places, which meant if we wanted to change something like its volume it was very hard to do. I've extracted scenes for the CandyButton's ClickSound and HoverSound so that this can be changed in one place.

### DIFF
--- a/project/src/main/ui/candy-button/HoverSound.tscn
+++ b/project/src/main/ui/candy-button/HoverSound.tscn
@@ -1,0 +1,9 @@
+[gd_scene load_steps=3 format=2]
+
+[ext_resource path="res://src/main/utils/DeconflictedAudioStreamPlayer.tscn" type="PackedScene" id=1]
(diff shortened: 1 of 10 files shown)


project/src/main/ui/candy-button/HoverSound.tscn
```diff
@@ -0,0 +1,12 @@
...
+
+[ext_resource path="res://src/main/utils/DeconflictedAudioStreamPlayer.tscn" type="PackedScene" id=1]
+[ext_resource path="res://assets/main/ui/button-hover.wav" type="AudioStream" id=2]
+
+[node name="HoverSound" instance=ExtResource( 1 )]
+pause_mode = 2
+stream = ExtResource( 2 )
+volume_db = -12.0
+__meta__ = {
+"_editor_description_": ""
```
Poobslag, line 11: unnecessary metadata

Poobslag: Removed

PR #2595, opened 2024-08-12
Prevent unbinding UI controls.

Poobslag:
UI controls can no longer be unassigned. If a control is unassigned, it is swapped with the input currently being rebound.

Added missing 'next_tab', 'prev_tab' ui actions

### DIFF
--- a/project/src/main/settings/keybind-settings.gd
+++ b/project/src/main/settings/keybind-settings.gd
@@ -70,6 +70,22 @@ func set_preset(new_preset: int) -> void:
 	emit_signal("settings_changed")
 
 
+## Restore the default keybind for a specific action.
+##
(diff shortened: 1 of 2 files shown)


project/src/main/settings/keybind-settings.gd
```diff
@@ -70,6 +70,16 @@ func set_preset(new_preset: int) -> void:
 	emit_signal("settings_changed")
 
 
+func restore_default_keybinds(action_name: String) -> void:
```
Poobslag, line 73: Add comment

Poobslag: Added comment, `## Restore the default keybind for a specific action.`

PR #2602, opened 2024-08-16
Redesigned TrainingMenu with LevelButtonScroller

Poobslag:
The training menu previously used an undecorated button which was very wide, and would have been difficult to reskin as a candy button given its width.

The training menu now uses a 'LevelButtonScroller' which uses the same button style as the career mode and level picker. We'd like to eventually amp up these buttons to include some icons or fancy graphics, so that will help this screen look a bit prettier too.

The training menu's high score table now uses the 'squeak-theme-gy' variant, instead of an undecorated 'h4' theme.

### DIFF
--- a/project/project.godot
+++ b/project/project.godot
@@ -674,6 +674,11 @@ _global_script_classes=[ {
 "language": "GDScript",
 "path": "res://src/main/world/letter-shooter.gd"
 }, {
+"base": "Control",
+"class": "LevelButtonScroller",
+"language": "GDScript",
+"path": "res://src/main/ui/level-select/level-button-scroller.gd"
+}, {
 "base": "Node",
 "class": "LevelEditor",
 "language": "GDScript",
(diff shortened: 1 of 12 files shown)


project/src/main/ui/candy-button/CandyButtonCollapsible.tscn
```diff
@@ -216,10 +216,10 @@ anchor_left = 0.5
...
-margin_top = -25.0
-margin_right = 25.0
-margin_bottom = 25.0
+margin_left = 95.0
+margin_top = 71.0
+margin_right = 145.0
+margin_bottom = 121.0
 rect_min_size = Vector2( 50, 50 )
 size_flags_vertical = 4
 expand = true
```
Poobslag, line 225: rv

Poobslag: Reverted.

PR #2620, opened 2024-08-23
Fixed TrainingMenu bug when selecting first level

Poobslag:
When selecting the first level in a region, the TrainingMenu's levelbuttonscroller would sometimes behave strangely and not scroll the button to the center.


project/src/main/ui/level-select/level-button-scroller.gd
```diff
@@ -94,9 +94,10 @@ func populate(new_region: Object, default_level_id: String = "") -> void:
 		_level_ids.append_array(_region.level_ids)
 
 	_refresh()
+	yield(get_tree(), "idle_frame")
 	if default_level_id:
-		yield(get_tree(), "idle_frame")
 		set_central_button_index(_level_ids.find(default_level_id))
+	_refresh_central_button_index(false)
```
Poobslag, line 100: Add a comment for this; it's a bit strange

Poobslag: Added comment, `set_central_button_index will only refresh the central button index ...`

PR #2639, opened 2024-08-27
Improve 'ui_cancel' behavior with modal dialogs.

### LINKED ISSUE #1787: Pressing escape closes settings menu, even when a dialog is visible
image
image

After showing a warning dialog, the player can press "escape" to close the settings menu (!)

This isn't something malicious the player would do -- pressing escape should just close the settings menu. Presumably this is the combination of two or three related bugs which don't just affect this dialog, but affect a lot of other dialogs and UI elements as well
  * dialogs should be modal, and shouldn't let you interact with the rest of the application
  * dialogs should react to the 'esc' or 'back' buttons

Poobslag:
'ui_cancel' now closes modal dialogs.

'ui_cancel' no longer closes the settings menu if a modal dialog is visible.

Closes #1787.

### DIFF
--- a/project/src/main/ui/settings/settings-dialogs.gd
+++ b/project/src/main/ui/settings/settings-dialogs.gd
@@ -13,6 +13,12 @@ signal delete_confirmed
 ## emitted when the player is prompted to delete a save slot and answers 'no' or closes the window
 signal delete_cancelled
 
+## emitted when a dialog becomes visible/invisible
+##
+## Parameters:
+## 	'value': 'true' if any dialog is visible, 'false' if all dialogs are invisible
+signal dialog_visibility_changed(value)
+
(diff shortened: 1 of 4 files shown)


project/src/main/ui/settings/settings-dialogs.gd
```diff
@@ -13,6 +13,9 @@ signal delete_confirmed
 ## emitted when the player is prompted to delete a save slot and answers 'no' or closes the window
 signal delete_cancelled
 
+## emitted when a dialog becomes visible/invisible
+signal dialog_visibility_changed(value)
```
Poobslag, line 17: This is worth a comment; its behavior is somewhat surprising; as 'value' could be the dialog

PR #2642, opened 2024-08-30
Added support for 'level icons'.

Poobslag:
Levels can now define an 'icons' property. Icons summarize a level's gimmicks, such as whether they have carrots or sharks, whether they affect the piece rotation or include tetrominos.

Added new 'smooth outline shader'.


project/src/main/ui/level-select/HardcoreLevelSelectButton.tscn
```diff
@@ -8,6 +8,19 @@
...
+border_width_left = 4
+border_width_top = 4
+border_width_right = 4
+border_width_bottom = 4
+border_color = Color( 0.423529, 0.262745, 0.192157, 1 )
+corner_radius_top_left = 8
+corner_radius_top_right = 8
+corner_radius_bottom_right = 8
+corner_radius_bottom_left = 8
+
```
Poobslag, line 23: This looks like it is repeating a resource in LevelSelectButton.tscn

Poobslag: Reverted; HardcoreLevelSelectButton and LevelSelectButton do not need to change

PR #2664, opened 2024-09-13
New receipt results screen with bar graph.

Poobslag:
Added a new results screen with a bar graph, custom animations and sound effects. A perfect score gets the player a rainbow-colored 'M' stamp, whereas a failing score gets the player a turd-colored duck stamp.


project/src/main/puzzle/receipt-stamp.gd
```diff
@@ -0,0 +1,53 @@
+extends Control
```
Poobslag, line 1: Should these scripts go in the 'receipt' folder?

Poobslag: Moved to `project/src/main/puzzle/result/receipt-stamp.gd`

PR #2669, opened 2024-09-16
Added chalk texture to chalkboards

Poobslag:
This affects the progress board, the victory screen chalkboard, and the chalkboard showing the player's score in puzzle mode.


Poobslag:
Created a chalkboard texture using Clip Studio Paint's Perlin Noise feature.

PR #2695, opened 2024-09-27
Removed legacy rank/duration calculator, show_x_rank fields

Poobslag:
These mechanisms are preserved on the 'legacy rank calculator' branch if for some reason we want to resurrect them.


project/project.godot
```diff
@@ -1751,6 +1731,7 @@ _global_script_class_icons={
 "CutsceneQueue": "",
 "CutsceneSearchFlags": "",
 "CutsceneWorld": "",
+"DemoSave": "",
```
Poobslag, line 1734: Remove this.

Poobslag: Removed.

PR #2705, opened 2024-10-01
Placeholder spanish localization

Poobslag:


project/assets/main/locale/es.po
```diff
@@ -12636,7 +12636,7 @@ msgstr "Cara"
 
 #: project/assets/main/locale/localizables-extracted.py:2653
 msgid "Eyes"
-msgstr ""
+msgstr "Ojos"
```
Poobslag, line 12639: Can we flag these translations as "requiring human review"?

Poobslag: I've added the `#, fuzzy` flag which appears to be the standard way of marking something as "a human needs to take action"

PR #2725, opened 2024-10-04
Add sprint belly animation work

va-0:
This is the animation work necessary for having a belly in the sprinting animation.
There's some code stuff you would need to do:
- Need to somehow set rgb_palette shader parameters. I tried "set_krgb" on the belly on creature load and it didn't work.
- Need to scale belly based on creature fatness. It's very easy; you just tune the scale of "SprintBellyScaleControl" in CreatureVisuals.tscn. Good-looking scale is 0 for no belly, 0.094 for a small belly, and .165 for a large belly
- Need to hide belly when not sprinting, then show on sprint animation. For now it's always visible.

_(Optional: There's just one animation frame of sprint-nw where the far leg is in front of the belly. It's barely noticable because the silhouette reads, but the fix would be moving that far leg to another texture and putting it on a lower Z layer. It still looks good even without fixing that.)_


Poobslag:
https://github.com/user-attachments/assets/79909304-5861-4d97-b465-b42d63a16099

https://github.com/user-attachments/assets/8bac0f80-ed66-4775-aa37-77665e334b61

Here's a preview of what the animation looks like, for anyone else looking at this pull request. Really interesting look! I'm not sure I'll merge this into the main game but it's amazing to see what people come up with.

Poobslag:
I'm personally not interested in merging this pull request at the moment, as it doesn't quite match my desired aesthetics for the game. Thanks for wrestling with my spaghetti code to make this cool demo though! It looks really interesting, even if it's not quite to my taste.

PR #2757, opened 2024-10-09
Added level intro goal message, 'Earn ¥25 with 15 lines...'

Poobslag:
Added a goal message which shows up before non-tutorial levels. This level serves two purposes. For normal levels, it gives the player an idea of what score they should be aiming for either for a baseline rank or to reach the next rank milestone. For boss levels, it tells the player the necessary criteria to progress.


project/src/main/puzzle/result/BarGraphGoal.tscn
```diff
@@ -36,5 +36,4 @@ margin_right = 198.0
 margin_bottom = 10.0
 custom_colors/font_color = Color( 0.305882, 0.227451, 0.196078, 1 )
 custom_fonts/font = SubResource( 2 )
-text = "¥25,000: SSS"
```
Poobslag, line 39: rv

Poobslag: Reverted

PR #2765, opened 2024-10-10
Boss levels show 'GOAL' marker on result screen.

Poobslag:
The result screen now includes a 'GOAL' marker for boss levels so the player can see how close they are to passing the level. When replaying the levels in training mode, the traditional rank markers are shown.


Poobslag:
image

The goal is shown as "GOAL" in all upper-case letters. We should localize it.

PR #2768, opened 2024-10-10
Glow effect for focused candy buttons.

Poobslag:
Candy buttons are very brightly colored, making it difficult to tell which one is currently focused. Even though it's a bright cyan color, this doesn't contrast significantly with bright yellow or bright red buttons, particularly for people who are colorblind.

I've added a pulsing animation to help the focused button stand out.

I gave CandyButtonDemo an opaque ColorRect background to reproduce layering problems which can arise in the game. I wanted this glow to go behind the button, but it's very difficult to do so. A naive approach is to set its Z-Index to 1 but this causes problems, which were not visible in CandyButtonDemo because it lacked a background.


Poobslag:
Old:
image

New:
image

There is now a pulsing blue glow effect. I also made the text outline a little brighter.

PR #2772, opened 2024-10-10
Analog sticks now work for UI sliders.

Poobslag:
The UI sliders have some specific logic for processing JoypadButton events, but this logic was being sidestepped because our AnalogEventTranslator generated InputEventActions instead. I've rewritten the translator to use InputEventJoypadButtons instead, which also makes the code much simpler.

### DIFF
--- a/project/src/main/ui/analog-event-translator.gd
+++ b/project/src/main/ui/analog-event-translator.gd
@@ -7,8 +7,8 @@ extends Node
 ## 	1. Pressing a direction on the analog stick activates the action many times, not just once
 ## 	2. Holding the direction 100% on the analog stick stops activating the action at all
 ##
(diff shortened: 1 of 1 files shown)


project/src/main/ui/analog-event-translator.gd
```diff
@@ -23,13 +23,8 @@ var _down_pressed := false
 var _left_pressed := false
 var _right_pressed := false
 
-## Joypad action mappings populated by scanning the InputMap.
```
Poobslag, line 26: This change in functionality should result in the header comment on Line 1 changing as well.

Poobslag: Changed comment to `sends out JoypadButton events...`

PR #2779, opened 2024-10-11
Add a 'controller disconnected' message during puzzles

Poobslag:
If the player's controller disconnects during gameplay, we pop up the settings menu with a message.

We don't pop up the settings menu in the creature editor, during cutscenes, or in other parts of the game.

### DIFF
--- a/project/src/main/ui/settings/settings-menu.gd
+++ b/project/src/main/ui/settings/settings-menu.gd
@@ -25,6 +25,8 @@ export (QuitType) var quit_type: int setget set_quit_type
 var _post_save_method: String
 var _post_save_args_array: Array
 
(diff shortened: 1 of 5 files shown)


Poobslag:
image

project/src/main/ui/settings/settings-menu.gd
```diff
@@ -60,10 +62,16 @@ func show() -> void:
 	_touch_buttons.visible = true
 	_dialogs.visible = true
 	_window.show()
+
```
Poobslag, line 65: rv

Poobslag: Reverted.

PR #2783, opened 2024-10-11
Placeholder spanish localization from machine translation.

Poobslag:


project/assets/main/locale/es.po
```diff
@@ -17283,6 +17284,7 @@ msgstr "Selección de Nivel"
 
 #: project/src/main/ui/menu/SystemButtons.tscn
 msgid "Credits"
+git status
```
Poobslag, line 17287: rv

Poobslag: Reverted.

PR #2788, opened 2024-10-13
Added BlocksDuringRules.top_out_effect. Added 'none' top out effect.

Poobslag:
Combined 'clear_on_top_out' and 'refresh_on_top_out' into a single 'top_out_effect' field. These two fields shouldn't have been separated, since otherwise it's ambiguous what should happen if both are set to true.

Added a new 'top_out_effect none'. Some tutorials don't want to alter the playfield when the player tops out, and just want to restart the tutorial section.


Poobslag:
This pull request also accidentally applied a fix for tutorial modules not resetting the `lost` and `topping_out` state, causing softlocks.

PR #2860, opened 2024-10-24
Added 'troubleshoot.bat' file, verbose logging

Poobslag:
Turbo Fat now includes a 'turbofat-troubleshoot.bat' file which enables Godot's '--verbose' flag as well as a Turbo-Fat-specific '--tf-verbose' flag. These enable enhanced logging for scene transitions, singletons and resource loading. The intent is to give players a tool to share detailed crash information with developers, to try and root out the cause of the game crashing after puzzles.

### DIFF
--- a/package.sh
+++ b/package.sh
@@ -17,6 +17,14 @@ echo "version=$version"
 
 win_export_path="project/export/windows"
 win_zip_filename="$win_export_path/turbofat-win-v$version.zip"
+win_bat_filename="$win_export_path/turbofat-win-troubleshoot-v$version.bat"
+win_bat_template="bin/troubleshoot_bat/turbofat-troubleshoot.bat.template"
+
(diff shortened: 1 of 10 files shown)


package.sh
```diff
@@ -17,16 +17,28 @@ echo "version=$version"
 
 win_export_path="project/export/windows"
 win_zip_filename="$win_export_path/turbofat-win-v$version.zip"
+win_bat_filename="$win_export_path/turbofat-win-troubleshoot-v$version.bat"
+template_filename="bin/troubleshoot_bat/turbofat-troubleshoot.bat.template"
```
Poobslag, line 21: Should this be called something like `bat_template_filename`? We have export templates, and this could be ambiguous.

Poobslag: Changed to `win_bat_template`.

PR #2868, opened 2024-10-25
Fixed SettingsMenu focus neighbours.

Poobslag:
The SettingsMenu has up to two quit buttons. Before, only the top 'quit' button had its focus neighbours set. Now, the focus neighbours are set for the topmost node which is visible in the tree, and the focus neighbours are unset for the other node.

The 'Customize Controls' tab has three topmost buttons. Before, focus neighbors were only configured for an arbitrary one of the three. Now, focus neighbors are configured for all three. Navigating down from the TabContainer selects the currently pressed button.

### DIFF
--- a/project/src/main/ui/settings/SettingsMenu.tscn
+++ b/project/src/main/ui/settings/SettingsMenu.tscn
@@ -158,7 +158,7 @@ size_flags_vertical = 3
 follow_focus = true
 
 [node name="VBoxContainer" type="VBoxContainer" parent="Window/UiArea/TabContainer/SoundAndGraphics"]
-margin_right = 712.0
+margin_right = 736.0
 margin_bottom = 306.0
 size_flags_horizontal = 3
 
@@ -582,6 +582,7 @@ margin_left = 198.0
 margin_right = 298.0
 margin_bottom = 29.0
 rect_min_size = Vector2( 100, 0 )
+focus_neighbour_top = NodePath("../../..")
 toggle_mode = true
 pressed = true
 group = SubResource( 3 )
@@ -592,6 +593,7 @@ margin_left = 318.0
 margin_right = 418.0
 margin_bottom = 29.0
 rect_min_size = Vector2( 100, 0 )
+focus_neighbour_top = NodePath("../../..")
 toggle_mode = true
 group = SubResource( 3 )
 text = "WASD"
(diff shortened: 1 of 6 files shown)


project/src/main/ui/settings/keybind/settings-keybinds-control.gd
```diff
@@ -165,22 +167,36 @@ func _assign_wasd_joypad_values() -> void:
 	_preset_control("SwapHoldPiece").visible = SystemData.gameplay_settings.hold_piece
 
 
+func _refresh_focus_neighbours() -> void:
```
Poobslag, line 170: Comment

project/src/main/ui/settings/settings-menu-bottom.gd
```diff
@@ -63,6 +66,8 @@ func _refresh_quit_type() -> void:
 		$HBoxContainer/VBoxContainer1/Holder1.visible = false
 		$HBoxContainer/VBoxContainer2/Spacer.visible = false
 		$HBoxContainer/VBoxContainer3/Spacer.visible = false
+	
+	_tab_input_enabler.refresh_focus_neighbours_for_current_tab()
```
Poobslag, line 70: Worth commenting this line, as it's a bit strange

project/src/main/ui/settings/SettingsMenu.tscn
```diff
@@ -118,7 +118,6 @@ valign = 1
 autowrap = true
 
 [node name="Window" type="Panel" parent="."]
-visible = false
```
Poobslag, line 121: rv

project/src/main/ui/settings/keybind/settings-keybinds-control.gd
Poobslag: Added comment: `Assigns the TabContainer's bottom focus neighbor to the pressed preset button.`

project/src/main/ui/settings/settings-menu-bottom.gd
Poobslag: Added comment, ` 	# refresh focus neighbors for the TabContainer and newly visible buttons`

project/src/main/ui/settings/SettingsMenu.tscn
Poobslag: Reverted

PR #2879, opened 2024-11-04
Added 'give up' confirmation for Adventure mode

### LINKED ISSUE #2847: Better warning when quitting/restarting Adventure Mode level
When quitting/restarting an Adventure Mode level for the first time, the player should be shown a popup warning "Giving up on this level will end your day in Adventure Mode! Are you sure you want to give up?"

Restarting should pop up the Settings Menu with the same pop-up.

If the player accepts, the popup should never be shown again.

The player should be able to re-enable this popup with some sort of "Re-enable Tutorials" button in the Settings menu.

Poobslag:
Before, hitting 'restart' or 'give up' would surprise or frustrate players, because Adventure mode would suddenly end and they might not realize what they had done wrong.

Now, the player is prompted '...will end the adventure day, do you really give up?'. This prompt can be restored with a button in the settings menu.

Closes #2847.

### DIFF
--- a/project/src/main/ui/settings/settings-menu.gd
+++ b/project/src/main/ui/settings/settings-menu.gd
@@ -6,6 +6,7 @@ extends CanvasLayer
 
 signal show
 signal hide
+signal give_up_confirmed
 signal quit_pressed
 signal other_quit_pressed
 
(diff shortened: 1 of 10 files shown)


project/src/main/ui/settings/settings-menu.gd
```diff
@@ -126,20 +134,18 @@ func _on_Bottom_ok_pressed() -> void:
 
 
 func _on_Bottom_quit_pressed() -> void:
-	# keep the menu shown, but unpause the scene tree
-	Pauser.toggle_pause("settings-menu", false)
-	
-	if quit_type in [SAVE_AND_QUIT, SAVE_AND_QUIT_OR_GIVE_UP]:
```
Poobslag, line 132: Let's merge this 'unpause the scene tree' logic separately.

Poobslag: Merged in #2880 .

PR #2891, opened 2024-11-07
Streamlined combo tutorial: failure_count, cakes_built

Poobslag:
Replaced 'failed_section' variable with 'failure_count' logic. This allows for more fine-grained logic depending on how many times the player has failed.

Reset 'level_performance.pieces' and 'cakes_built' in one place, instead of specifying it for each tutorial section. This allows all tutorial sections to have fine-grained logic depending on how many pieces the player has placed.


Poobslag:
This pull request also tweaks the first part of the tutorial so it starts the player with a combo.

This prevents the player from dropping two pieces off to the side right away. It's not something most players would do anyways, but I think it makes sense to treat it as a failure.

PR #2904, opened 2024-11-19
Difficulty menu is forced on the player once.

Poobslag:
After completing the tutorial, the player is forcibly shown the difficulty menu. This ensures all players have the option to select an easier difficulty, even if they don't notice the main menu button. This also ensures that if they later encounter trouble, they'll hopefully remember 'Oh, I chose my difficulty earlier, I should lower it.'

If the player clicks 'Re-enable tips', the player is also shown the tutorial after the splash screen.


Poobslag:
image

When the difficulty menu is forced upon the player, we have a little tip at the top 'Customize your difficulty' to give the player context.

PR #2906, opened 2024-11-19
Persist save data after upgrading it.

Poobslag:
Before, after upgrading old save data, the old data would remain on disk and need to be upgraded every time the player saved the game. This isn't a big deal, since newer versions are written if the player makes any change or plays any level. But it's possible a player could configure their settings, and not touch their settings for 6 months, in which case the old settings version would stay on their disk until they adjusted the volume or switched their save slot.

Save data is now immediately persisted after it is loaded.

### DIFF
--- a/project/src/main/data/player-save.gd
+++ b/project/src/main/data/player-save.gd
@@ -38,6 +38,9 @@ var legacy_filename := "user://turbofat0.save" setget set_legacy_filename
 ## 'true' if rank data should be populated, and invalid levels purged from the player's save. Can be changed for tests
 var populate_rank_data := true
 
+## 'true' if the previous '_save_items_from_file' call upgraded the loaded data.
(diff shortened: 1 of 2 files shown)


project/src/main/data/player-save.gd
```diff
@@ -38,6 +38,9 @@ var legacy_filename := "user://turbofat0.save" setget set_legacy_filename
 ## 'true' if rank data should be populated, and invalid levels purged from the player's save. Can be changed for tests
 var populate_rank_data := true
 
+## 'true' if the previous 'load_player_data_from_file' call upgraded the loaded data.
```
Poobslag, line 41: `_save_items_from_file`; not `load_player_data_from_file`

Poobslag: Changed to `_save_items_from_file`

PR #2927, opened 2024-11-26
Tomatoes indicate lines which can't be cleared.

Poobslag:
Before, some levels like 'Slow Service' wouldn't clear lines until a delay, but players often found this frustrating or confusing, or even thought it was a bug.

Now, tomatoes show up and indicate the lines which can't be cleared. They count down with their fingers to let the player know when they'll go away.


Poobslag:
Snapshot

Tomatoes now show up holding out their hand like "nuh-uh"! The row is also blocked out with a red shadow, and there are some clock particles.

PR #2944, opened 2024-12-11
Skipping dialog skips choices as well.

### LINKED ISSUE #1007: More efficient dialog skipping
Skipping through dialog efficiently currently involves holding down the "skip" key, and then pressing/reholding it when prompted for dialog options. However this first press is often ignored, as a protection against players who repeatedly mash the "next line" button.

This first press should not be ignored if the player was just skipping dialog.

Poobslag:
I don't think people would usually skip the dialog but then care about making a specific choice. It's a little tedious having to hold a key to skip dialog, then let go and repress it during a dialogue choice, sometimes multiple times if you trigger the "don't mash through dialog choices" prevention code.

Closes #1007.

### DIFF
--- a/project/src/main/chat/ui/touch-translator.gd
+++ b/project/src/main/chat/ui/touch-translator.gd
@@ -7,6 +7,9 @@ export (NodePath) var narration_frame_path: NodePath
 ## index of the current touch event, or -1 if there is none
 var _touch_index := -1
 
+## 'true' if the player is making a dialog choice
+var _showing_choices := false
+
 ## scancode which triggers a ui_accept action.
 ## echo events cannot be emitted without an InputEventKey instance which requires a scancode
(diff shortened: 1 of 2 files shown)


project/src/main/chat/ui/touch-translator.gd
```diff
@@ -7,6 +7,9 @@ export (NodePath) var narration_frame_path: NodePath
 ## index of the current touch event, or -1 if there is none
 var _touch_index := -1
 
+## 'true' if the player is making a dialog choice
+var _showing_choices := false
```
Poobslag, line 11: This field is never used.

Poobslag: Unfortunately, it's used on line 76 which was already merged (and which does not compile without this change, oops.)

PR #2946, opened 2024-12-17
Decreased difficulty of "Fruit Salad"

Poobslag:


Poobslag:
This level was basically impossible to complete at the speeds expected of players in the "Kflab" region (20-30 pieces per minute). It inserted so many extra lines that you had to play pretty quickly to keep up. I've decreased the rate of the extra lines.

I experimented with an alternate approach which inserted lines every 5 lines you cleared, removing the time component entirely -- but a playtester didn't seem to find that approach any easier.

PR #2950, opened 2024-12-17
Separate 'M' and 'TOP' grades, to allow for lenient master grade

Poobslag:
RankCriteria now defines a distinct 'TOP' grade. Currently this is the exact same as the 'M' grade, but we can now decrease the score/time requirements for 'M' grades to make the grade more achieveable.

Since levels define their own arbitrary values for what constitutes an 'M' rank, this might be counterintuitive. Why define an arbitrary value of what a 'TOP' grade is, only to have master grades be an arbitrary fraction of it? However, keeping these separate allows us to more easily make sweeping changes such as deciding a master grades is 99% or 80% of a perfect grade. Otherwise, we'd need to make those changes in 200 different level definitions.

### DIFF
--- a/project/src/main/puzzle/rank-criteria.gd
+++ b/project/src/main/puzzle/rank-criteria.gd
@@ -4,6 +4,10 @@ class_name RankCriteria
 ## Each level defines the rank criteria for the best rank, and can optionally define rank criteria for other ranks.
 ## The remaining ranks are calculated by interpolation.
 
+## The 'TOP' grade corresponds to TAS-level play. It is never awarded to the player, but it is defined in our level
+## files and other grades are extrapolated from it.
+const PERFECT_GRADE := "TOP"
+
 ## Each level defines the rank criteria for the best rank, but if no rank criteria is specified these values are used
 ## instead. These values are arbitrary and will not produce meaningful output, so we log a warning if they are used.
 const DEFAULT_BEST_SCORE := 2000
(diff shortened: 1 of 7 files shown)


project/src/main/puzzle/rank-criteria.gd
```diff
@@ -187,22 +193,22 @@ func copy_from(rank_criteria: RankCriteria) -> void:
...
 		return
 	
-	var new_threshold: int = thresholds_by_grade[Ranks.BEST_GRADE]
+	var new_threshold: int = thresholds_by_grade[PERFECT_GRADE]
 	if duration_criteria:
 		new_threshold = new_threshold * TIME_PERCENT_BY_GRADE[grade] / factor
 	else:
 		new_threshold = new_threshold * SCORE_PERCENT_BY_GRADE[grade] * factor
-	new_threshold = _sanitize_threshold(new_threshold)
+	new_threshold = sanitize_threshold(new_threshold)
```
Poobslag, line 205: Merge this 'sanitize_threshold' change separately.

project/src/main/world/environment/lemon/lemon-obstacle-library.tres
```diff
@@ -470,6 +470,7 @@ points = PoolVector2Array( -3, 80, 29, 64, 93, 96, 61, 112 )
 4/autotile/navpoly_map = [  ]
 4/autotile/priority_map = [  ]
 4/autotile/z_index_map = [  ]
+4/autotile/fallback_mode = 0
```
Poobslag, line 473: rv

project/src/main/puzzle/rank-criteria.gd
Poobslag: Reverted.

project/src/main/world/environment/lemon/lemon-obstacle-library.tres
Poobslag: Reverted.

PR #2977, opened 2025-01-03
Fixed unusual bug which let pieces warp with flip inputs

Poobslag:
The logic for caching a piece's previous location for flip inputs did not check whether the cached piece was the same as the flipped piece. This meant you could execute some very specific and strange inputs to warp a the flipped piece to the previous piece's location.

### DIFF
--- a/project/src/main/puzzle/piece/piece-rotator.gd
+++ b/project/src/main/puzzle/piece/piece-rotator.gd
@@ -24,8 +24,11 @@ signal rotated_180(piece)
 
 export (NodePath) var input_path: NodePath
 
+## piece information used to restore the its previous location when flipping it
 var _prev_rotate_frames := 0
+var _prev_rotate_piece: ActivePiece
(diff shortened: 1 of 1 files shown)


project/src/main/puzzle/piece/piece-rotator.gd
```diff
@@ -213,7 +218,9 @@ func _calc_rotate_target(piece: ActivePiece) -> void:
 ## kick it out of a nook, so the piece must be flipped without first being rotated. This function moves the piece back
 ## to where it was before the player held the first rotate button.
 func _restore_prerotate_pos(piece: ActivePiece) -> void:
-	if _prev_rotate_frames > FLIP_FRAME_LENIENCY:
+	if _prev_rotate_piece != piece:
+		pass
```
Poobslag, line 222: Add a comment

Poobslag: Comment: `not the same piece; don't let players ...`

PR #2982, opened 2025-01-10
Fixed sporadic 'class instance is gone' errors in CandyButton

Poobslag:


Poobslag:
These errors started coming up occasionally while testing the creature editor.

PR #2983, opened 2025-01-10
Bug: Creature editor shows black pixels when selecting feature

### LINKED ISSUE #2531: Creature Editor should not show black pixels when selecting feature
When changing the creature's appearance in the creature editor, parts of the creature blink black as the creature is redrawn. We should avoid this.

One approach might be to have an offscreen creature where we make the changes, and then we swap them in once the appearance is updated.

Another approach might be to replace the creature with a poof cloud which gradually clears once the appearance is updated.

Or, maybe a little combination of both.

Poobslag:
When changing the creature's appearance in the creature editor, parts of the creature blink black as the creature is redrawn.

To avoid this, I've introduced an offscreen creature where we apply the changes, and then we swap them in once their apperance is updated.

Closes #2531.

### DIFF
--- a/project/src/main/editor/creature/allele-buttons.gd
+++ b/project/src/main/editor/creature/allele-buttons.gd
@@ -49,6 +49,15 @@ onready var _overworld_environment: OverworldEnvironment = get_node(overworld_en
 onready var _creature_saver: CreatureSaver = get_node(creature_saver_path)
 onready var _creature_editor_library := Global.get_creature_editor_library()
 
+func _ready() -> void:
+	for creature in [_player(), _player_swap()]:
+		# Disable physics processing for both the player and their doppelganger. Creatures are KinematicBody2D
+		# instances, and their physics interferes with the swapping process.
+		creature.set_physics_process(false)
+		
+		creature.connect("dna_loaded", self, "_on_Creature_dna_loaded", [creature])
+
+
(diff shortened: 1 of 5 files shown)


Poobslag:
While the black pixels no longer appear, it might be undesirable that there's so much lag when selecting a new feature. We could reduce this lag with a more complex algorithm which, instead of regenerating the creature's entire appearance, only changes one particular feature or one shader.

project/src/main/ui/candy-button/candy-button-h3.gd
```diff
@@ -152,6 +152,21 @@ func _initialize_onready_variables() -> void:
...
+	_hover_sound.pitch_scale = rand_range(0.95, 1.05)
+	SfxKeeper.copy(_hover_sound).play()
+
+
+func _apply_mouse_exited_effects() -> void:
+	if get_tree().is_connected("idle_frame", self, "_apply_mouse_exited_effects"):
+		get_tree().disconnect("idle_frame", self, "_apply_mouse_exited_effects")
+	emit_signal("hovered_changed")
+
+
```
Poobslag, line 169: These changes should have been merged separately.

Poobslag: Rebased branch.

PR #2995, opened 2025-02-06
Support threaded saving to eliminate lag.

Poobslag:
PlayerSave.save_player_data() now supports optional threaded saving, reducing lag from 50-100ms to 0ms by moving the operation to a secondary thread.

This improves responsiveness but sacrifices automicity, so it is not used universally.

### DIFF
--- a/project/src/main/data/player-save.gd
+++ b/project/src/main/data/player-save.gd
@@ -47,11 +47,21 @@ var save_scheduled := false
 ## Provides backwards compatibility with older save formats
 var _upgrader := PlayerSaveUpgrader.new().new_save_item_upgrader()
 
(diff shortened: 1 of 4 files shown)


project/src/main/data/player-save.gd
```diff
@@ -47,13 +47,20 @@ var save_scheduled := false
 ## Provides backwards compatibility with older save formats
 var _upgrader := PlayerSaveUpgrader.new().new_save_item_upgrader()
 
+var _save_thread: Thread = null
```
Poobslag, line 50: Document this field.

Poobslag: Added a comment, `The currently active thread which is saving the player's data.`

PR #2997, opened 2025-02-07
Touch size settings: tick count, limits

Poobslag:
Decreased the maximum touch size from 1.69 to 1.40. In an upcoming change which introduces the hold piece, a scale of 1.69 is too large and causes the buttons to overlap.

Settings menu no longer hard-codes the max value and tick count, these values are derived from the SettingsTouchSize.VALUES constant.

### DIFF
--- a/project/src/main/ui/settings/SettingsMenu.tscn
+++ b/project/src/main/ui/settings/SettingsMenu.tscn
@@ -856,9 +856,7 @@ margin_bottom = 19.0
 rect_min_size = Vector2( 80, 16 )
 size_flags_horizontal = 3
 size_flags_vertical = 4
-max_value = 11.0
 value = 1.0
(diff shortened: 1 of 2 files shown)


project/src/main/ui/settings/SettingsMenu.tscn
```diff
@@ -858,7 +858,7 @@ size_flags_horizontal = 3
 size_flags_vertical = 4
 max_value = 11.0
```
Poobslag, line 859: max_value shouldn't be 11.0; it should either be the correct value (10) or it should be unset

Poobslag: Removed; max_value and tick_count have their defaults in the Godot Editor, and are only assigned when running the game.

